PR #55, opened 2017-07-27
Remove application and model from bundles. Added plugin reference.

stefanoborini:
Removed application and model from the bundles. The application is still available by traversing the plugin, which is now available and initialized. The model must be passed appropriately to the methods of the created classes.


force_bdss/data_sources/base_data_source_bundle.py
```diff
@@ -18,8 +19,14 @@ class BaseDataSourceBundle(ABCHasStrictTraits):
     #: A human readable name of the bundle. Spaces allowed
     name = String()
 
+    plugin = Instance(Plugin)
```
martinRenou: Can you put this in the interface as well ? Like you did for the `i_mco_bundle` and `i_kpi_calculator_bundle`

PR #60, opened 2017-07-31
Moved parameters into MCO

stefanoborini:
Moved the parameters from generic area to being MCO specific


force_bdss/io/workflow_reader.py
```diff
@@ -128,6 +120,7 @@ def _extract_mco(self, wf_data):
         mco_bundle = registry.mco_bundle_by_id(mco_id)
         model_data = wf_data["mco"]["model_data"]
         model_data["parameters"] = self._extract_mco_parameters(
+            mco_id,
```
martinRenou: When using the method `_extract_mco_parameters` you already know about the mco bundle that you are using. So instead of giving the id we could pass the mco bundle itself and it will be faster to find the mco parameter factories later with `mco_parameter_factory_by_id(mco_bundle, factory_id)`

PR #68, opened 2017-08-01
Introduced data value object

stefanoborini:
Replaced previous system of Result and Parameter objects for MCO, DataSource and KPICalculator with a set of DataValue objects that contain all the relevant information for a specific entity created by the MCO/DS/KPI


force_bdss/core_evaluation_driver.py
```diff
@@ -31,18 +31,18 @@ def application_started(self):
...
             data_source = ds_bundle.create_data_source()
-            ds_results.append(data_source.run(ds_model, parameters))
+            ds_results.extend(data_source.run(ds_model, mco_data_values))
 
         kpi_results = []
         for kpic_model in workflow.kpi_calculators:
             kpic_bundle = kpic_model.bundle
             kpi_calculator = kpic_bundle.create_kpi_calculator()
-            kpi_results.append(kpi_calculator.run(kpic_model, ds_results))
+            kpi_results.extend(kpi_calculator.run(kpic_model, ds_results))
```
martinRenou: You are not passing the DataValues coming from the MCO. Will this come later with the slots ?

force_bdss/core/data_value.py
```diff
@@ -0,0 +1,34 @@
+from traits.api import HasStrictTraits, Any, String, Enum
+
+
+class DataValue(HasStrictTraits):
```
martinRenou: Should we specify where does the DataValue come from ? Like a `from` trait:
```
from = Enum('MCO', 'MODEL', 'DATASOURCE', 'KPI')
```
Or will this be part of the slots ?

stefanoborini: I thought about it, we used to have originator, but I start to believe it's not useful. we can reintroduce it at a later stage with no fuss, because the setting will be done by the core anyway, so plugin authors don't have to do that manually, and nothing will change for them.

force_bdss/core_evaluation_driver.py
stefanoborini: That's right. at the moment they are not passed. They will be passed later, we I actually resolve the names, and pass only what is resolved.

force_bdss/core/data_value.py
martinRenou: I think it can be useful for the GUI, when displaying the results in the table we want to be able to distinguish KPIs to other parameters. But I agree that it's better if the author of the plugin don't have to deal with this.

stefanoborini: In the gui you will not have access to it anyway. The cli will send you a message with the relevant information. You will never have to deal with DataValue objects.

martinRenou: Ok

stefanoborini:
@martinRenou good to merge?

martinRenou:
@stefanoborini It's fine for me

PR #69, opened 2017-08-02
Introduce slots and resolution of named variables

stefanoborini:
Introduces the concept of slots to data sources, and the application of named variables to each slot when running


force_bdss/core_evaluation_driver.py
```diff
@@ -31,18 +31,96 @@ def application_started(self):
...
+
+    def _bind_data_values(self,
+                          available_data_values,
+                          model_slot_map,
+                          slots):
+
+        passed_data_values = []
+        lookup_map = {dv.name: dv for dv in available_data_values}
+
+        if len(slots) > len(model_slot_map):
```
martinRenou: Even if in the case of `len(slots) < len(model_slot_map)` it won't result in a failure, we should consider it a FileError, no ?

martinRenou: if `if len(slots) < len(model_slot_map)` it also means that there is an issue in the file, no ?

stefanoborini: I remember I had a valid reason for this, but I can't recall it now, so I guess you are right.

force_bdss/core_evaluation_driver.py
```diff
@@ -31,18 +33,194 @@ def application_started(self):
...
+
+                logging.error(error_txt)
+                raise RuntimeError(error_txt)
+
+            if len(res) != len(ds_model.output_slot_names):
+                error_txt = (
+                    "The number of data values ({} values) returned"
+                    " by the DataSource '{}' does not match the number"
+                    " of user-defined names specified ({} values)."
+                    " This is likely a DataSource plugin error.").format(
```
martinRenou: `This is either an input file error or a plugin error` ?

force_bdss/core_plugins/dummy/kpi_adder/kpi_adder_calculator.py
```diff
@@ -16,3 +17,15 @@ def run(self, model, data_source_results):
                 type=model.cuba_type_out,
                 value=sum
             )]
+
+    def slots(self, model):
+        return (
+            (
+                Slot(type=model.cuba_type_in),
```
martinRenou: We could already have a dynamic number of slots, depending on the model ? For the KPIAdder we could have a `model.number_inputs` integer

stefanoborini: I am planning to support that via variadic slots.

PR #76, opened 2017-08-08
Adds change_slots event to notify UI of slot-modifying changes

### LINKED ISSUE #74: Add changes_slots event to the base model class.
We want an event we can listen on so that when triggered it will force recomputing of the slots according to the current options. This will then be used by the UI to present the correct slots.

The user is responsible for triggering this event in the plugin model he writes. We expect this to be relatively uncommon, but a possible use case for those data sources that actually do change slots (both input and output) depending on some model values (e.g. configuration options).

stefanoborini:
When the model contains options that might modify the resulting slots,
a notification must be issued, so that the UI can take appropriate action.

Provides that mechanism, and documents it.

Fixes #74


stefanoborini:
@martinRenou done

PR #97, opened 2017-08-22
Add CodeEditor datasource [DO NOT MERGE]

martinRenou:
This data source allows the user to type python code or load an external python script which define his data source.

This branch must NOT be merged in `force_bdss` master. It will eventually be added to the Enthought plugins when those plugins will be extracted from this repository. It depends on `traitsui` and `pyface`, which shouldn't be the case of the `force_bdss`.
This PR is only here for showing that some work has been done on a CodeEditor plugin, and that it should be moved later to an other repository.


stefanoborini:
Closing this one. It probably is sound but I don't have time to check it.

PR #101, opened 2017-08-24
Create probe classes for tests

martinRenou:


force_bdss/mco/base_mco_communicator.py
```diff
@@ -23,6 +23,8 @@ def __init__(self, factory, *args, **kwargs):
         self.factory = factory
         super(BaseMCOCommunicator, self).__init__(*args, **kwargs)
 
+        super(BaseMCOCommunicator, self).__init__(*args, **kwargs)
```
stefanoborini: this is no longer needed.

martinRenou: Oh sorry, maybe an problem when merged

PR #108, opened 2018-04-23
Introduced log file option

stefanoborini:
Allows to write a log file instead of relying to stdout/err.


force_bdss/bdss_application.py
```diff
@@ -65,7 +67,8 @@ def _load_failure_callback(plugins, manager, entry_point, exception):
     Reports failure to load a module through stevedore, using the
     on_load_failure_callback option.
     """
-    logging.error(
+    log.error(
```
dpinte: if you want you can combine the `log.error` and `log.exception` by adding a kwarg `exc_info=True` to `log.error`.

PR #148, opened 2018-06-07
[WIP] 144 add dev docs

jjenthought:
Added install instructions which include dev/doc requirements.


jjenthought:
just wanting to check the update to .travis.yml is ok, i.e nothing relies on that specfic edm version or anything like that

stefanoborini:
Ok, this one is now outdated as we use the ci package to handle it, so I close it.

Issue #134, opened 2018-05-18
traits_version still present in output_slot_info and input_slot_info


stefanoborini:
The traits_version still appears in workflow.json input_slot_info and output_slot_info. It should be removed from the serialization.

stefanoborini:
WorkflowWriter needs to be modified appropriately.

PR #152, opened 2018-06-15
Removed __traits_version__ from json files

jjenthought:
The workflow is a dictionary which contains lists of dictionaries, so only the __traits_version__ keys in the top level dict were being removed. This changed the function so it should do it recursively.
It seems fine but the only test is a on a single hardcoded example at the minute. Refs. issue https://github.com/force-h2020/force-bdss/issues/134

### DIFF
--- a/force_bdss/io/workflow_writer.py
+++ b/force_bdss/io/workflow_writer.py
@@ -93,29 +93,30 @@ def traits_to_dict(traits_obj):
     """Converts a traits class into a dict, removing the pesky
     traits version."""
 
-    def pop_traits(dictionary):
-        """Recursively remove the __traits_version__ attribute
-        from dictionary."""
-        try:
-            dictionary.pop("__traits_version__")
-        except KeyError:
-            pass
-
-        for key in dictionary:
-            # If we have a dict, remove the traits version
-            if isinstance(dictionary[key], dict):
-                pop_traits(dictionary[key])
-            # If we have a non-dict which contains a dict, remove traits from
-            # that as well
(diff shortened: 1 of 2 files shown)


force_bdss/io/workflow_writer.py
```diff
@@ -91,10 +92,31 @@ def _model_data(self, model):
...
+    """Recursively remove a named key from dictionary and any contained
+    dictionaries."""
     try:
-        state.pop("__traits_version__")
+        dictionary.pop(remove_key)
     except KeyError:
         pass
 
-    return state
+    for key in dictionary:
```
stefanoborini: for key, value in dictionary.items()

force_bdss/io/workflow_writer.py
```diff
@@ -91,10 +92,31 @@ def _model_data(self, model):
...
+        # If remove_key is in the dict, remove it
+        if isinstance(dictionary[key], dict):
```
stefanoborini: use value here

force_bdss/io/workflow_writer.py
```diff
@@ -91,10 +92,31 @@ def _model_data(self, model):
...
+            pop_recursive(dictionary[key], remove_key)
+        # If we have a non-dict iterable which contains a dict,
+        # call pop.(remove_key) from that as well
+        elif isinstance(dictionary[key], Iterable):
```
stefanoborini: a string is also an iterable. Just check if it's list or tuple.

stefanoborini:
flake

PR #157, opened 2018-06-20
Added an empty list return from parameter factories

stefanoborini:
Small fix to return an empty list if the parameter_factories is undeclared. The default implementation should just return an empty list.


stefanoborini:
I can cover it, I just left it pending

PR #163, opened 2018-06-29
Verifier Additions

jjenthought:
Just adds a check for kpis, since they got moved to the mco. Also adds the layer a datasource occupies as part of the error message


force_bdss/core/verifier.py
```diff
@@ -38,13 +36,26 @@ def _check_mco(workflow):
...
-                                        error="Parameter has empty type"))
+                                        error="Parameter {} "
+                                              "has empty type".format(idx)))
+
+    for idx, kpi in enumerate(mco.kpis):
+        if len(kpi.name.strip()) == 0:
+            errors.append(VerifierError(subject=kpi,
+                                        error="KPI {} has empty name".format(
+                                            idx)))
+        if len(kpi.objective.strip()) == 0:
```
stefanoborini: This is an enumeration, it can never be empty unless we add the empty string as an entry in the enum itself.

force_bdss/core/verifier.py
```diff
@@ -62,18 +73,19 @@ def _check_execution_layers(workflow):
...
 
         for ds in layer.data_sources:
-            errors.extend(_check_data_source(ds))
+            errors.extend(_check_data_source(ds, idx))
 
     return errors
 
 
-def _check_data_source(data_source_model):
+def _check_data_source(data_source_model, layer_no):
```
stefanoborini: layer_number

force_bdss/core/verifier.py
```diff
@@ -100,23 +112,27 @@ def _check_data_source(data_source_model):
     if len(input_slots) != len(data_source_model.input_slot_info):
         errors.append(VerifierError(
             subject=data_source_model,
-            error="Missing input slot name assignment"))
+            error="Missing input slot name assignment "
+                  "in layer {}".format(layer_no)))
```
stefanoborini: you can also put the data source name here for additional information. You get it from the factory

force_bdss/core/verifier.py
jjenthought: I think with KPIs the name is initially set to be empty, or if you remove the datasource providing the output designated as a KPI.
Similarly for the kpi objective I thought it made sense to have the empty string as a default option, as a guard against someone running a simulation with a viable but unhelpful objective. e.g. minimising when they needed to maximise. I once wasted ~8 days of compute time because I did something like this!

stefanoborini: ok, so what you need is Enum('', "MINIMISE"). The first argument is always the default. You don't need to treat it as a string.

force_bdss/core/verifier.py
```diff
@@ -38,13 +36,27 @@ def _check_mco(workflow):
...
-                                        error="Parameter has empty type"))
+                                        error="Parameter {} ({}) has empty "
+                                              "type".format(idx, p_name)))
+
+    for idx, kpi in enumerate(mco.kpis):
+        if len(kpi.name.strip()) == 0:
+            errors.append(VerifierError(subject=kpi,
+                                        error="KPI {} has empty name".format(
+                                            idx)))
+        if kpi.objective == '':
```
stefanoborini: This technically cannot happen unless we add the empty option to the objective. I am conflicted if we should provide one or not, you made a good point about adding it though...
I'd suggest to add it to the enumeration as well, in first position.

stefanoborini:
Minor thing and coverage as a consequence about the option of having '' as a kpi option. I'd say go for it.

force_bdss/core/verifier.py
```diff
@@ -38,13 +36,27 @@ def _check_mco(workflow):
...
+                                        error="Empty Name - Parameter {} "
+                                              "({})".format(idx, p_desc)))
         if len(param.type.strip()) == 0:
             errors.append(VerifierError(subject=param,
-                                        error="Parameter has empty type"))
+                                        error="Empty Type - Parameter {} "
+                                              "({})".format(idx, p_desc)))
+
+    for idx, kpi in enumerate(mco.kpis):
+        if len(kpi.name.strip()) == 0:
```
stefanoborini: The strip is not needed because it's an Identifier.

force_bdss/core/verifier.py
```diff
@@ -100,23 +113,31 @@ def _check_data_source(data_source_model):
...
 
     if len(output_slots) != len(data_source_model.output_slot_info):
         errors.append(VerifierError(
             subject=data_source_model,
-            error="Missing output slot name assignment"))
+            error="Missing output slot name assignment "
+                  "in layer {}".format(layer_number)))
 
     for idx, info in enumerate(data_source_model.output_slot_info):
         if len(info.name.strip()) == 0:
```
stefanoborini: can remove the strip

PR #166, opened 2018-07-03
Changes slots now uses traits metadata instead of requiring setting the event

### LINKED ISSUE #102: use traits metadata to indicate if a trait changes the slots.
Currently we require users to bind the traits that can change the slots to a handler that invokes changes_slots event. We can achieve the same by requiring a metadata entry on those traits that change the slots, and implement the binding with the event in the base class.

stefanoborini:
Instead of having to attach a listener to traits that change the slots, we can simply give `changes_slots` metadata to the trait so that it gets registered automatically

Fixes #102


stefanoborini:
@jjenthought we would actually have to port all the plugins, but it's not a big deal, I think that the only one worth porting is the enthought example. I don't think you have anything in the wfmanager that will be influenced by this.

PR #171, opened 2018-07-12
Workflow Verification

jjenthought:
Required by https://github.com/force-h2020/force-wfmanager/pull/194 as some of the error reporting things are now in verifier.py. 
Changes the VerifierError object to have a local & global error message, which can be the same. The global message should be something useful when looking at the workflow as a whole, and the local one useful when the object causing the error is selected/ being edited.

### DIFF
--- a/force_bdss/core/verifier.py
+++ b/force_bdss/core/verifier.py
@@ -1,4 +1,7 @@
 import logging
+
+from itertools import groupby
+
 from traits.api import HasStrictTraits, Str, Any
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


force_bdss/core/verifier.py
```diff
@@ -116,31 +146,74 @@ def _check_data_source(data_source_model, layer_number):
...
+    consecutively wherever possible.
+    For example an input of [0,1,2,4,5,7] returns the string '0-2, 4-5, 7' """
+    index_list.sort()
+    # Single, consecutive or non-consecutive
+    if len(index_list) == 1:
+        return str(index_list[0])
+    else:
+        repl = []
+
+        for i, index_group in groupby(enumerate(index_list), lambda val:
```
stefanoborini: be wary of using a plain i. It's really hard to grep for.

force_bdss/core/verifier.py
```diff
@@ -116,31 +146,74 @@ def _check_data_source(data_source_model, layer_number):
...
+
+def multi_error_format(index_list):
+    """Takes a list of integers and returns a string where they are grouped
+    consecutively wherever possible.
+    For example an input of [0,1,2,4,5,7] returns the string '0-2, 4-5, 7' """
+    index_list.sort()
+    # Single, consecutive or non-consecutive
+    if len(index_list) == 1:
+        return str(index_list[0])
+    else:
```
stefanoborini: you don't need the else clause here. you are returning in the first case. Keeps the code less indented.

stefanoborini:
So I guess we will need this one to support the other changes.

PR #174, opened 2018-07-16
Added name, version and description to plugins

stefanoborini:
Adds name version and description to plugins so that the information can be used by the upcoming plugin manager.
We must update all plugins to provide this information.

### DIFF
--- a/force_bdss/base_extension_plugin.py
+++ b/force_bdss/base_extension_plugin.py
@@ -147,7 +147,7 @@ class BaseExtensionPlugin(Plugin):
                 self.__class__))
 
     def get_description(self):
-        return u"No description available"
+        return "No description available"
 
     def get_factory_classes(self):
         """Must return a list of factory classes that this plugin exports.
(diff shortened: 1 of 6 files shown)


stefanoborini:
@jjenthought I need your review on this one. Thanks

force_bdss/base_extension_plugin.py
```diff
@@ -105,12 +127,28 @@ def __init__(self, *args, **kwargs):
             self._logger.exception(e)
             self.error_msg = str(e)
             self.error_tb = traceback.format_exc()
+            self.name = ""
```
jjenthought: Would there be any difference between u"" and ""? I'd imagine not since it's just an empty string but I don't know how traits checks these things

stefanoborini: With the specification of unicode_literals now this string is a unicode string. I am going to introduce the same future import in the rest of the patch.

PR #176, opened 2018-07-23
Verifier Fix

jjenthought:
The verifier now warns when all outputs are unassigned in a datasource. While this is likely to change in future, it's more reasonable then what we have now.
Before it was warning only when a single output was missing, which was a mistake on my part when setting it up.

### DIFF
--- a/force_bdss/core/verifier.py
+++ b/force_bdss/core/verifier.py
@@ -11,7 +11,7 @@ class VerifierError(HasStrictTraits):
     subject = Any()
     #: An error message relevant to the local modelview
     local_error = Str()
-    #: An error message relevent to the overall workflow
+    #: An error message relevant to the overall workflow
     global_error = Str()
 
     def __init__(self, subject, global_error='', local_error=''):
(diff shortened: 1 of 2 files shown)


force_bdss/core/verifier.py
```diff
@@ -178,16 +182,18 @@ def _check_data_source(data_source_model, layer_number):
             global_error="Missing output slot name assignment "
                          "in layer {}".format(layer_number)))
 
+    #: Check if any datasources have all outputs unnamed
     row_index_errors = []
```
stefanoborini: I would change this logic to check only if all are undefined. Or as we said, the possibility that all could be undefined is not zero, although unlikely.

jjenthought:
So after we spoke my thought was: if you have a datasource in the workflow just for a side-effect it’s reasonably painless to just go and name one of the output parameters, even if you don’t want to use it in the end. However it would be quite annoying to run a simulation having forgotten to name something that you did want to keep track of. But that’s all pretty subjective!

Were you thinking of having it so you only get an error if all outputs in all datasources are undefined? At the minute it gives errors if all outputs in a single datasource are undefined.

James

stefanoborini:
Good point. So let's keep it as "at least one". We'll have to add a "trash" datasource later.

stefanoborini:
In any case, the current logic can just check for the names, not get the errors and check that

PR #180, opened 2018-07-25
Added the install step to the CI package, instead of having it in the build-env

stefanoborini:


stefanoborini:
@jjenthought yes it does, thanks

stefanoborini:
Built on top of #179

PR #181, opened 2018-07-25
Removed force-bootstrap from the travis environment. using default

stefanoborini:
Staying true to the documentation, we use the default environment to setup the bootstrap


stefanoborini:
@jjenthought ok, I'll fix it, one sec

PR #187, opened 2018-07-31
Propagate data values in MCO events

stefanoborini:
- Restructures events so that we propagate the full DataValue, rather than the plain values.
- Changes the naming of entities on the events.
- makes started and finished events obsolete.

This is a backward incompatible change. wfmanager and plugins must be readapted.


stefanoborini:
@jjenthought this one is ready

PR #204, opened 2019-03-29
Fix order of KPIs returned from execute_workflow

### LINKED ISSUE #203: MCOs can return `optimal_kpis` that do not match KPISpecification
See force-h2020/force-wfmanager#259. I think issue comes down to the bdss. Plugins should not have to manually adhere to the `KPISpecification` in `BaseMCOModel`, instead it should be handled either by the `BaseMCO` class (as a wrapper to notify_new_point) or through the `execute_workflow` function.

ml-evs:
Fixes #203 and force-h2020/force-wfmanager#259.

`execute_workflow` now returns KPI's exactly as specified by name in the KPISpecification, however there are a couple of things to think about:

- Currently KPISpecification can contain duplicates. These will now be handled correctly by `execute_workflow` (i.e. returned), but it doesn't really make sense to allow this.
- The KPISpecification can contain names that aren't bound to any output slots (though the wfmanager doesn't allow this); it might be worth raising an error if this happens.
- KPISpecification can be empty (even using wfmanager). This doesn't cause the BDSS any issue, but most plugins assume that they are getting a non-empty list back, and often set e.g. weights based on the length of this list.

### DIFF
--- a/force_bdss/core/execution.py
+++ b/force_bdss/core/execution.py
@@ -34,8 +34,9 @@ def execute_workflow(workflow, data_values):
 
     kpi_results = [
         dv
+        for kpi_name in kpi_names
         for dv in available_data_values
-        if dv.name in kpi_names
+        if dv.name == kpi_name
     ]
 
     return kpi_results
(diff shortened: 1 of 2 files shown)


corranwebster:
LGTM.

PR #205, opened 2019-04-05
Move verification to class that is being verified.

corranwebster:
This permits subclasses to add to the verification methods (important for plugin makers).  It's also useful for GUIs as they can more easily access the verification for just the instance being inspected in a particular UI.

Verification is implemented via a method, so there should be no overhead for running code unless it _wants_ to perform verification.


force_bdss/core/verifier.py
```diff
@@ -2,195 +2,43 @@
...
 
-        super(VerifierError, self).__init__(
-            subject=subject,
-            global_error=global_error,
-            local_error=local_error
-        )
+    #: The object that has an error.
+    subject = Any
+
+    #: An optional trait that holds the error value.
```
jjenthought: Could we extend the docstring here to explain that this refers to a trait name within `subject`. At first I wasn't sure what a error value was until seeing it in use

PR #206, opened 2019-04-05
Fix Qt application instantiation

corranwebster:
Because we look at all plugins, we potentially do TraitsUI imports which in turn potentially select a ETS toolkit.  This forces early ETS toolkit selection to `null` in the `BDSSApplication` which should prevent the Qt application being instantiated with all the overhead that goes with that.

The result is a substantial speedup in the command-line app.

### DIFF
--- a/force_bdss/bdss_application.py
+++ b/force_bdss/bdss_application.py
@@ -16,13 +16,6 @@ from .core_mco_driver import CoreMCODriver
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


corranwebster:
I think this is the best I can do on test coverage.

jjenthought:
The coverage thing is a bit of a pain for small PRs, I tried do find a way to disable it for PRs <100 lines but it doesn't seem to be easily doable.

PR #217, opened 2019-05-15
Added look up of custom data views when loading plugins

ml-evs:
Required for #force-h2020/force-wfmanager#276.


ml-evs:
Not sure what's going on with the Travis PR build, but this should be a trivial merge... (unless you want to go a different route for plugins contributing data views).

nicolasap-dm:
It would be better to avoid calls to `get_data_views` from `bdss`, so that the plugins can "hide" the imports of the graphical stack within that function, and only ever exercise those imports if run from a GUI.

nicolasap-dm:
PR in `force_wfmanager` refactored so that il will be the one that calls `get_data_views`; this avoids graphical imports in headless runs.

PR #221, opened 2019-06-10
Update installation.rst

jjenthought:
Just some updates to make the install process a little more clear, as it relies on people entering commands while in the correct directory


nicolasap-dm:
As discussed offline with @jjenthought, I'll add a commit to this PR to move from `pushd`/`popd` to `cd` which is arguably more known

jjenthought:
Yeah looks good - it also reminded me to put in a bit saying you can load a workflow file from the command line in the workflow manager! I'll merge this as soon as it lets me

PR #222, opened 2019-06-13
Upgrade dependencies

nicolasap-dm:
Upgrading all dependencies to the latest served by EDM/PyPI. 

Flake8 becomes a bit more restrictive.

This doesn't seem to cause test failures down the line (`force-wfmanager` and `force-bdss-plugin-enthought-example`), except some flake8 errors, fixing in https://github.com/force-h2020/force-wfmanager/pull/286.

### DIFF
--- a/ci/__main__.py
+++ b/ci/__main__.py
@@ -5,8 +5,8 @@ DEFAULT_PYTHON_VERSION = "3.6"
 PYTHON_VERSIONS = ["3.6"]
 
 CORE_DEPS = [
-    "envisage==4.7.1-1",
-    "click==6.7-1",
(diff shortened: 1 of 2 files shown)


nicolasap-dm:
Note: due to a possible bug in EDM (https://github.com/enthought/edm/issues/2038) the environment here will contain pyface 6.0.0 (while it could, in theory, use pyface 6.1.0). We like that, because `force-wfmanager` is not ready for pyface 6.1.0, but it's rather random and we should switch to a bundle-based `ci` task in the future.

corranwebster:
LGTM

PR #223, opened 2019-06-21
CI updates, including macOS worker and verbose tests

nicolasap-dm:
This PR:
- adds a macOS Travis CI job
- uses EDM 1.11.0 in CI
- uses a Python 3.6 bootstrap environment (was 3.5)

Also:
- makes the tests verbose

I'm not sure why `language: c` and `ccache` were used (probably boilerplate?), but I removed them in favor of `language: generic`.

Codecov is run only once (on Linux) since it would be redundant otherwise.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,33 @@
 language: c
+sudo: false
+
+env:
+  global:
+    - EDM_FULL=1.11.0
+    - EDM_X_Y=1.11
+
 cache:
   directories:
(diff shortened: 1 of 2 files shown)


flongford:
@nicolasap-dm Is this ready to go?

nicolasap-dm:
Yes, I wouldn't have merged without review, although this repository allows it

flongford:
Cool, LGTM

PR #224, opened 2019-06-21
Bundle based environment building

nicolasap-dm:
With this PR, the process for building an environment doesn't change (from the user point of view): indeed, `.travis.yml` and `docs/source/installation.rst` aren't changed.

However, now the dependencies are not specified in the `ci` code, but in bundle files, which makes the environment creation process entirely reproducible.

---

When updating the dependencies, instead of changing the `*_DEPS` constants, we run
```
ci generate-edm-bundles  # this will pick up the newest eggs
```
then, if
```
ci build-env
ci install
ci test
```
work well, we `commit` and `push` the new bundles.

If downgrades are necessary, then and only then we edit `EDM_DEPS` with version strings on selected packages, and repeat the process.

### DIFF
--- a/ci/__main__.py
+++ b/ci/__main__.py
@@ -1,22 +1,30 @@
 import click
-from subprocess import check_call
+from distutils.version import StrictVersion
+from subprocess import check_call, check_output
+import os
 
 DEFAULT_PYTHON_VERSION = "3.6"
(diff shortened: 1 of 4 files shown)


nicolasap-dm:
Very likely out of date. Closing.

PR #225, opened 2019-07-16
Docs: instructions for contributing data_views in the Plugin Development chapter

nicolasap-dm:
In order to advertise the feature being added in wfmanager (https://github.com/force-h2020/force-wfmanager/pull/276).

It doesn't reflect anything that is present in this codebase, but this repo holds the documentation for writing plugins so it seemed like a good place to put it.

### DIFF
--- a/doc/source/plugin_development.rst
+++ b/doc/source/plugin_development.rst
@@ -248,6 +248,10 @@ them discoverable by a GUI application (wfmanager) through the
         of its subclasses.
         """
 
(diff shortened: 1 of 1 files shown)


flongford:
Agreed, that's the file I was thinking about.

Maybe comment that the `get_data_views` method does not exist in the `BaseExtensionPlugin` and so is not required to be overridden. Just to make it clear that it is an optional extra and users don't need to write their own data views.

PR #226, opened 2019-07-22
Including verification check for no defined KPIs in BaseMCOModel

flongford:
This PR completes the verification requirements for `BaseMCOModel`, listed in `<BaseMCOModel.verify>`:

    Check that the MCO model:
        - has at least one parameter
        - has at least one KPI
        - has no parameter errors
        - has no KPI errors

Specifically, the method now raises a `VerifierError` if no KPIs are defined in the MCO, replicating the condition if no parameters are present.

Supporting unit tests in `test_verifier.py` and `test_core_mco_driver.py` files have also been updated accordingly.

### DIFF
--- a/force_bdss/core/tests/test_verifier.py
+++ b/force_bdss/core/tests/test_verifier.py
@@ -30,9 +30,19 @@ class TestVerifier(unittest.TestCase):
         wf.mco = self.plugin.mco_factories[0].create_model()
 
         errors = verify_workflow(wf)
-        self.assertEqual(len(errors), 2)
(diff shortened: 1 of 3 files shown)


flongford:
@jjenthought - As a final check, do you know whether the at least one KPI verification check was missing for any reason?

jjenthought:
> @jjenthought - As a final check, do you know whether the at least one KPI verification check was not implemented for any reason?

I can't think of any good reason not to implement this, if you're trying to run something but not return any results it sounds like a user error. Which is especially useful to be warned about if your simulation would take days/weeks..

PR #230, opened 2019-09-25
ENH: Refactored BaseDataSourceModel slot synchronization from WfManager to BDSS 

### LINKED ISSUE #229: BaseDataSource and BaseDataSourceModel are entangled
Currently we are unable to determine all information regarding the nature of input and output slots for a data source without instantiating both `BaseDataSource` and `BaseDataSourceModel` classes for a given `BaseDataSourceFactory`.

Notably, the `slots` method on the `BaseDataSource` determines the number of input and output slots as well as their CUBA types. However, it can require information from the `BaseDataSourceModel` if the numbers of slots is dynamic (i.e. depends on user input). Additionally, the `BaseDataSourceModel` handles user input of variable names, and so even if the number of slots is fixed we need the model class to collate these.

Consequently, the WfManager has needed an extra `DataSourceView` class to combine both sets of information, which includes also creating instantiations of the `BaseDataSource` before the MCO is fired. This appears highly redundant, and has led to unnecessary complexity (force-h2020/force-wfmanager/issues/219, force-h2020/force-wfmanager/issues/246, force-h2020/force-wfmanager/issues/307, #213)

A possible solution would be to combine both classes, or at least move the `slots` method into the `BaseDataSourceModel`, so all information needed by the UI is stored in one object. However, this would critically alter the structure of all existing plugin data source classes, and so it should first of all be justified as a worthwhile change.

flongford:
This PR resolves #229 

Synchronisation of slot number and type occurs between `BaseDataSource` and `BaseDataSourceModel` classes upon initiation of a `BaseDataSourceModel`.

The private method `<object: BaseDataSourceModel>._sync_data_source_slots` carries out the following steps:

- A `BaseDataSource` object is instantiated, via the assigned `factory` trait (a instance of type `BaseDataSourceFactory`)
- Expected input and output slots are generated by calling the `slots` method and passing in the `BaseDataSourceModel` as the `model` argument.
- A check occurs between any assigned `input_slot_info` or `output_slot_info` traits on the `BaseDataSourceModel`, and the expected input and output slots. If the lengths of each do not match, a `RuntimeError` is raised.
- The `type` attribute on each `Slot` is synchronised with the corresponding `type` trait on each `InputSlotInfo` and `OutputSlotInfo` entry in `input_slot_info` and `output_slot_info` lists.


force_bdss/data_sources/base_data_source_factory.py
```diff
@@ -70,7 +70,21 @@ def create_data_source(self):
         BaseDataSource
             The specific instance of the generated DataSource
         """
-        return self.data_source_class(self)
+
+        try:
+            data_source = self.data_source_class(self)
```
flongford: These try statements were being used in multiple places upon calling `factory.data_source_class`, so it seemed sensible to include them inside the method.

force_bdss/data_sources/base_data_source_model.py
```diff
@@ -46,43 +46,218 @@ class BaseDataSourceModel(BaseModel):
     #: this and adapt the visual entries.
     changes_slots = Event()
 
-    def verify(self):
-        """ Verify the data source model.
+    def __init__(self, factory=None, input_slot_info=None,
```
flongford: Here we pass in all arguments except `input_slot_info` and `output_slot_info`, so that their default values will be assigned first. The `factory` trait is included explicitly only to allow for backwards compatibility, since it is often passed in as an unnamed first argument,

force_bdss/core/input_slot_info.py
```diff
@@ -22,6 +23,12 @@ class InputSlotInfo(HasStrictTraits):
     #: The user defined name of the variable containing the value.
     name = Identifier()
 
+    #: The CUBA key of the slot
+    type = CUBAType()
```
flongford: We include `type` and `description` traits in both `InputSlotInfo` and `OutputSlotInfo` so that they contain all information from their corresponding `Slot` classes that is required by the WfManager UI. By doing so, we only need to refer to one object further down the line, rather than two.

force_bdss/core/output_slot_info.py
```diff
@@ -16,6 +16,12 @@ class OutputSlotInfo(HasStrictTraits):
     #: The user defined name of the variable containing the value.
     name = Identifier()
 
+    #: The CUBA key of the slot
+    type = CUBAType()
+
```
flongford: Same as comment for `OutputSlotInfo`

force_bdss/data_sources/base_data_source.py
```diff
@@ -14,8 +14,8 @@ class BaseDataSource(ABCHasStrictTraits):
     factory = Instance(IDataSourceFactory)
 
     def __init__(self, factory, *args, **kwargs):
-        self.factory = factory
         super(BaseDataSource, self).__init__(*args, **kwargs)
+        self.factory = factory
```
flongford: Assignment of trait has been moved until after super class initiation.

force_bdss/data_sources/base_data_source_model.py
```diff
@@ -46,43 +46,218 @@ class BaseDataSourceModel(BaseModel):
...
+            x.__getstate__() for x in self.output_slot_info
+        ]
+
+        return state
+
+    # -------------------
+    #   Private Methods
+    # -------------------
+
+    def _generate_slots(self):
```
flongford: This method is essentially lifted from the `DataSourceView` class in `force_wfmanager`, it creates an instance of the `BaseDataSource` class assigned to the `factory` trait and uses its `slots` method to generate the expected format of the `input_slot_info` and `output_slot_info` attributes.

force_bdss/data_sources/base_data_source_model.py
```diff
@@ -46,43 +46,218 @@ class BaseDataSourceModel(BaseModel):
...
+
+    def _reset_data_source_slots(self):
+        """Reset input_slot_info and output_slot_info attributes to
+        their defaults, based on the return value of a BaseDataSource
+        slots method"""
+
+        self.input_slot_info = self._input_slot_info_default()
+        self.output_slot_info = self._output_slot_info_default()
+
+    def _sync_slot_info(self, new_info, old_info):
```
flongford: The method provides a way to sync the `name` attribute between two `InputSlotInfo` or `OutputSlotInfo` objects, with some extra type checking.

force_bdss/data_sources/base_data_source_model.py
```diff
@@ -46,43 +47,187 @@ class BaseDataSourceModel(BaseModel):
...
-            logger.exception(
-                "Unable to create data source from factory '%s', plugin "
-                "'%s'. This might indicate a  programming error",
-                self.factory.id,
-                self.factory.plugin.id,
+    # -------------------
+    #      Defaults
+    # -------------------
+
+    def _input_slot_info_default(self):
```
flongford: The default values for `input_slot_info` and `output_slot_info` are generated from the associated `BaseDataSource` of the `factory` attribute. Importantly, we assign the corresponding `type` and `description` attributes on each `InputSlotInfo` and `OutputSlotInfo`, since we can use these as references that should not be changed by the user.

force_bdss/data_sources/base_data_source_model.py
```diff
@@ -46,43 +47,187 @@ class BaseDataSourceModel(BaseModel):
...
+                new_info, old_info, 'name',
+                attr_checks=['__class__', 'type', 'description'],
+                ignore_default=True
+            )
+
+    # -------------------
+    #   Public Methods
+    # -------------------
+
+    def slot_info_defaults(self):
```
flongford: Here we just provide a public method to obtain the default values of `input_slot_info` and `output_slot_info`, as they will be needed in `force_wfanager` after a `change_slot` event occurs

force_bdss/data_sources/base_data_source_model.py
```diff
@@ -46,43 +47,187 @@ class BaseDataSourceModel(BaseModel):
...
+                " created by factory {}, plugin {}. This might "
                 "indicate a programming error.",
                 self.factory.id,
                 self.factory.plugin.id
             )
             raise
 
+        return input_slots, output_slots
+
+    def _assign_slot_info(self, name, new_slot_info):
```
flongford: This private method is only expected to be used upon initiation, if arguments for `input_slot_info` or `output_slot_info` have been provided. For example, upon loading in of a workflow file.

In which case, it is necessary to combine any slot information in the workflow file with the default values, and therefore the number of input and output slots must match. When loading a workflow file, the `name` attribute on each slot is always defined by the user, but the `type` and `description` attributes must match those generated by the associated `BaseDataSource`.

force_bdss/data_sources/base_data_source_model.py
dpinte: I wonder if  you would  not want to make the initialization explicit in place of relying on the trait defaults. Then you would call `_generate_slots`  only once and do the initialization of both  inputs and outputs at the same time.

force_bdss/data_sources/data_source_utilities.py
```diff
@@ -0,0 +1,121 @@
...
+
+    source_attr = getattr(source, name)
+    target_attr = getattr(target, name)
+
+    # If ignore_default perform a check to see whether attribute on
+    # source is its default value, and skip cross check if so
+    if ignore_default:
+        try:
+            if source_attr == source.trait(name).default:
+                return True
```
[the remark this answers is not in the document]
flongford: Yes, that's confusing I agree. This block should probably be refactored into a separate test, otherwise the logic of the outcome is lost.

force_bdss/core/execution.py
```diff
@@ -64,16 +64,9 @@ def execute_layer(layer, environment_data_values):
...
-            log.exception(
-                "Unable to create data source from factory '{}' "
-                "in plugin '{}'. This may indicate a programming "
-                "error in the plugin".format(
-                    factory.id,
-                    factory.plugin.id))
-            raise
+
+        # Create an instance of the data source contained in the factory
+        data_source = factory.create_data_source()
```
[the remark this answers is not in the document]
flongford: If I remember correctly, the try/except block was moved to occur inside the `factory.create_data_source` method, since this just calls an additional abstract method on the factory, which would be the source of any `Exceptions`.

force_bdss/data_sources/base_data_source_model.py
flongford: I've now implemented a way that doesn't rely on trait defaults, meaning that now checks occur on the `input_slot_info` and `output_slot_info` attributes that have been assigned from the arguments passed into the `__init__` method.

force_bdss/data_sources/base_data_source_model.py
```diff
@@ -68,6 +68,50 @@ def _trigger_changes_slots(self, obj, name, new):
...
+                # changed 'type' and 'description' values
+                merge_lists(slot_info, attr, ['type', 'description'])
+            else:
+                # Otherwise update the list attributes by retaining any
+                # elements that have matching 'type' and 'description'
+                # values
+                new_list = retain_list(
+                    slot_info, attr, ['type', 'description']
+                )
+                setattr(self, attr_name, new_list)
```
Corwinpro: This is more of a comment on the `retain_list` and `merge_lists` behaviour: in one case we `setattr` implicitly, while for the other one we have to do it manually. Do I miss the design pattern here, or it could be possible to unify these two slot updates?

flongford: Yes, I agree this looks a bit strange. The `retain_list` function returns a new list that contains any elements of the old list (`attr` argument) that we want to keep.

A different design of `retain_list` would rather mutate `attr` in place, however since we are inserting or removing elements, this becomes difficult to do whilst iterating through them.

Corwinpro: What happens if instead of `merge_lists(slot_info, attr, ['type', 'description'])` we use
```
new_list = retain_list(
                    slot_info, attr, ['type', 'description']
                )
                setattr(self, attr_name, new_list)
```
?

flongford: Ah, yes we could do - actually that would be a more robust way since the `merge_lists` function assumes that the order of elements in each list is the same (they are referring to the same variables)

force_bdss/data_sources/base_data_source_model.py
```diff
@@ -68,6 +68,50 @@ def _trigger_changes_slots(self, obj, name, new):
         if changes_slots:
             self.changes_slots = True
 
+    @on_trait_change('changes_slots')
+    def _update_slot_info(self):
```
Corwinpro: Apparently, this method implements not only the `slot_info` update, but also the `slot` signature. The signature is changed when the `len(slot_info) != len(attr)`. Maybe there is a traits-way to handle these two different changes?

force_bdss/data_sources/data_source_utilities.py
```diff
@@ -188,3 +212,53 @@ def merge_lists_with_check(new_list, old_list, attributes):
...
+    -------
+    retained_list: list
+        List containing all elements to be retained
+    """
+    retained_list = []
+    old_list = copy.copy(old_list)
+
+    for new_element in new_list:
+        retained = False
+        for old_element in old_list:
```
Corwinpro: This refers to my previous comment in [here](https://github.com/force-h2020/force-bdss/pull/230/files#r345851967)

A possible way to make this more robust (this is not a working solution but just a sketch) is:
```
retained_list = []
    old_list = copy.copy(old_list)

    min_len = min(len(new_list), len(old_list))
    merge_lists(new_list[:min_len], old_list[:min_len], attributes)
    retained_list = old_list[:min_len]

    for new_element in new_list[min_len:]:
        retained = False
        for old_element in old_list[min_len:]:
```

PR #232, opened 2019-10-04
Refactor BaseFactory to not carry Envisage Plugin attribute

### LINKED ISSUE #209: Factory plugin trait seems to be excessive
As best I can tell, the `factory.plugin` trait is only used for extracting the `id` of the plugin (mainly for error reporting purposes, which is good) but carrying around the entire Envisage `plugin` seems excessive when perhaps just remembering `plugin_id` or other information may be enough.

### LINKED ISSUE #212: Evaluation and MCO runner data are entangled
Currently we have entanglement in the `Workflow` between information which is needed to evaluate at a data point (parameters, kpis, and execution layers) and information which is required to drive the MCO process (communicators, notifiers, ui_hooks, but also parameters and kpis).

It would be good to factor out the bits required for execution into an object (this is the primary thing being edited in the workflow manager GUI), the information needed for communicating with the MCO within an individual execution, and the information that the MCO runner needs.

This would be a fairly major design change, but would simplify a lot of code.  It should be done with care if it is done at all.

flongford:
This PR closes #209 and closes #212 

Specifically, the `plugin` attribute of type `<envisage.Plugin>` on `BaseFactory` class has been replaced by `plugin_id` and `plugin_name`, both of type `str`. Subsequent error messages and unit tests that reference `factory.plugin.id` have been updated to `factory.plugin_id`.

This change will lead to 2 breakage points in existing code:

1. Where a `BaseFactory` instance requires a `Plugin` argument to be passed into to the `init` method (i.e - everywhere)
2. Where any lower-level class is attempting to access attributes of the plugin via `self.factory.plugin.<attr>` (such as #209)
 
In order to maintain backwards compatibility for 1., the `plugin` argument in `BaseFactory.__init__` has been retained, with the possibility of it being of either `Plugin` type, or a dictionary containing keys `id` and `name`, referencing values to set to both `plugin_id` and `plugin_name` attributes. 

We shouldn't seek to directly support code that will be affected by 2., since we would like to remove this behaviour anyway. For example, any `BaseMCO` subclasses that reference the `Workflow` object through `factory.plugin.application.workflow` (i.e - all of them). The same goes for any references to `factory.plugin.application.workflow_file.path`.

A solution for this particular case is provided by creating a new class `WorkflowEvaluator`, that contains all "workflow" related information and can calculate the KPI values for a given state, defined by an input set of parameter values. This is achieved by creating an instance and calling `workflow_evaluator.evaluate(<list of parameters>)`, passing in a list of (basic python) values that correspond to each parameter. The same code for doing so was currently present in all `BaseMCO` subclasses in plugins, so this also seems like a very useful refactor in general.

An example of this functionality is provided in force-h2020/force-bdss-plugin-enthought-example/pull/34

### Change Log
- Envisage `plugin` attribute removed from `BaseFactory` class, though `plugin` argument is retained in class init method.
- New attributes `plugin_id` and `plugin_name` added to `BaseFactory`, which are a reference to the id and name of the Envisage plugin used to initiate the factory respectively.
- New interface classes `IEvaluator` and `WorkflowEvaluator` that are designed to act as a bridge between the application and the MCO, removing the need for the MCO to access plugin attributes
- `BaseMCO.run` method now takes in an initial argument `evaluator`, of type `IEvaluator`
- `OptimizeOperation` passes in a `WorkflowEvaluator` instance to each MCO run, rather than a `BaseMCOModel` instance

### Known Issues
~~- Subprocess mode of `WorkflowSolver` not fully tested in unit tests as cannot call `force_bdss` executable~~ 
_This functionality is now removed from the class_

~~- Naming of `ISolver` and `WorkflowSolver` may not be appropriate - they only are designed to return the state of the workflow at a single position in phase space.~~ 
_Renamed to `IEvaluator` and `WorkflowEvaluator`_

~~- Open question as to how to best pass new `WorkflowEvaluator` model into existing signature of `BaseMCO.run` without breaking current plugin subclasses that only expect one argument.~~ 
_This could be done via inspection of the method signature, but it doesn't avoid breaking backwards compatibility since all existing plugin `BaseMCO` classes will need updating anyway as they refer to `factory.plugin.application.workflow`, which no longer exists._


force_bdss/mco/i_evaluator.py
```diff
@@ -0,0 +1,35 @@
...
+
+class IEvaluator(Interface):
+    """Class that is used as a bridge between the BDSS application and
+    MCO runner. It can be passed in as an argument to a MCO run and
+    used to evaluate the state of a system for a given set of
+    parameters, returning a set of KPIs. This avoids the need for the
+    MCO to obtain any information regarding the Envisage application."""
+
+    #: An instance of the MCO model information
+    mco_model = Instance(BaseMCOModel)
```
flongford: This class is really just a container for a `BaseMCOModel` and any extra non-MCO specific information that could be used to evaluate the KPIs at a given data point. In our `BDSSApplication`, systems are described by `Workflow` objects, so we have also included a `WorkflowEvaluator` class that can evaluate the workflow KPIs.

force_bdss/app/workflow_evaluator.py
```diff
@@ -0,0 +1,65 @@
...
+            for parameter, value in zip(
+                self.mco_model.parameters, parameter_values)]
+
+        kpi_results = self.workflow.execute(data_values)
+
+        # Return just the values to the MCO, since the DataValue
+        # class is not specific to the BaseMCO classes
+        kpi_values = [kpi.value for kpi in kpi_results]
+
+        return kpi_values
```
flongford: This code was being implemented by every `BaseMCO` subclass in our plugins, so it made sense to refactor it into the core package as the standard way to evaluate the KPIs of a `Workflow` object

flongford:
As it's not possible to avoid breaking existing plugin `BaseMCO` classes, I'll prepare all the PRs requires to update each repo before merging to master

PR #234, opened 2019-10-31
[WIP] Refactor BaseDataSourceModel Slots generation

### LINKED ISSUE #229: BaseDataSource and BaseDataSourceModel are entangled

flongford:
This PR resolves #229 

Here we propose a way to simplify the slot interdependencies between a `BaseDataSource` and `BaseDataSourceModel`. This is achieved in a two-fold process:

1. Firstly by introducing a `BaseDataSourceModel._slot_info_generator` method for creating `input_slot_info` and `output_slot_info` attributes that depend on the return value of the associated `BaseDataSource.slots` method. By doing so, we are also able to obtain any `type` and `description` attributes that are defined for each slot directly from the `BaseDataSourceModel`, removing the need for any additional UI objects to interact with the `BaseDataSource`.

2. Secondly by enforcing a series of checks for any `input_slot_info` and `output_slot_info` arguments passed into the `BaseDataSourceModel.__init__` method, to ensure that they are in agreement with the expected values. These checks are performed in the `_assign_slot_info` method and include:

- The length of `input_slot_info` and `output_slot_info` lists must match the length of the respective lists returned by `_slot_info_generator`.
- Each element in the lists must possess the same `type` and `description` attribute values as their corresponding elements returned by `_slot_info_generator`. This is only ignored if any attributes are undeclared in the input arguments and are therefore set at their default values.

If either of these checks fails, a `RuntimeError` will be raised.

In the first instance, we would like to retain any `input_slot_info ` or `output_slot_info` elements referring to variables that have been defined before. These are identified as having matching `type` and `description` attributes with any elements in the new slot lists returned by `_slot_info_generator`, and are retained in the new values for `input_slot_info ` or `output_slot_info` (closing force-h2020/force-wfmanager/issues/51)

In the second instance, expect each element on the existing `input_slot_info` or `output_slot_info` lists to refer to the new lists returned by `_slot_info_generator`. Therefore we simply update any non-matching attribute `type` and `description` values between the corresponding elements.

### Notes

Another possible solution to minimise these dependencies would be to move the `slots` method from the `BaseDataSource` class onto the `BaseDataSourceModel` class. However, since `slots` is an abstract method that is designed to be implemented by plugin developers, we would still be left with the same verification requirements upon initiation of `input_slot_info` and `output_slot_info` (for example, when a workflow json file is parsed), as well as the need for a flexible update method to the UI. Additionally this solution would also require implementing workarounds to ensure backwards compatibility with all existing plugin `BaseDataSource` classes that possess a `slots` method.

A further option would be to not raise a `RuntimeError` upon initiation of `BaseDataSourceModel` in step 2., but simply overwrite any user input from a workflow json that did not contain the expected attribute values. However, this change should be communicated back to the user, and the most obvious way of doing so would be to raise an error during the construction of the `Workflow`.

## Change log
- `InputSlotInfo` and `OutputSlotInfo` classes contain two new attributes: `type` and `description`
- `BaseDataSourceModel` class contains new methods:  `_return_slots`, `_slot_info_generator`, `_assign_slot_info` and `_update_slot_info`, in order to apply process described above.
- New file `data_source_utilities.py` contains generic routines `merge_trait`, `trait_similarity_check`, `attr_checker`, `merge_trait_with_check`, `merge_list`, `merge_list_with_check` and `retain_list` for handling assignments and updates to `input_slot_info` and `output_slot_info` attributes.


force_bdss/data_sources/base_data_source_model.py
```diff
@@ -46,6 +50,148 @@ class BaseDataSourceModel(BaseModel):
...
+            x.__getstate__() for x in self.output_slot_info
+        ]
+
+        return state
+
+    # -------------------
+    #   Private Methods
+    # -------------------
+
+    def _return_slots(self):
```
Corwinpro: Would it be beneficial to change the name to this?
```
    def _data_source_slots(self):
```

Corwinpro:
Can we make the `InputSlotInfo` and `OutputSlotInfo` be instances (or inheriting) from a `BaseSlotInfo`? Right now it seems that we have two classes with very similar behaviour. If we could think of any other usage of the `BaseSlotInfo`, maybe it is worth thinking about refactoring it into a separate class.

flongford:
> Can we make the `InputSlotInfo` and `OutputSlotInfo` be instances (or inheriting) from a `BaseSlotInfo`? Right now it seems that we have two classes with very similar behaviour. If we could think of any other usage of the `BaseSlotInfo`, maybe it is worth thinking about refactoring it into a separate class.

Yes, that's a good idea indeed

force_bdss/data_sources/data_source_utilities.py
```diff
@@ -0,0 +1,196 @@
...
+        List of new elements
+    old_list: list
+        List of old elements to be retained
+    attributes: str or list of str, optional
+        An attribute or list of attributes to both merge and
+        check similarity on each corresponding element in
+        new_list and old_list
+    ignore_default: bool, optional, default: False
+        Whether or not to ignore any attributes on source or target
+        that are are equal to their default values
```
Corwinpro: `merge_lists_with_check` doesn't have the `ignore_default` parameter

force_bdss/data_sources/data_source_utilities.py
```diff
@@ -0,0 +1,196 @@
...
+    """
+
+    # Perform a merge_trait_with_check on each corresponding
+    # element of new_list and old_list
+    for new_element, old_element in zip(new_list, old_list):
+        for attr in attributes:
+            merge_trait_with_check(
+                new_element, old_element, attr, attributes,
+                ignore_default=True
```
Corwinpro: (Same comment) Perhaps this should be `ignore_default= ignore_default`

force_bdss/data_sources/data_source_utilities.py
flongford: Good spot

force_bdss/data_sources/base_data_source_model.py
```diff
@@ -46,27 +50,59 @@ class BaseDataSourceModel(BaseModel):
...
 
-        The data source model must have:
-        - input and output slots match between instance and model
-        - all output slots named
-        - no errors in input or output slots
+    # -------------------
+    #      Listeners
+    # -------------------
+
+    @on_trait_change("+changes_slots")
```
Corwinpro: I would appreciate if you could tell me how this works (and the `ChangesSlotsModel` in `test_base_data_source_model.py` as well). Thank you!

flongford: Sure, this `on_trait_change` handler listens to all traits on the `BaseDataSourceModel` class that have the `changes_slots=True` metadata and is called whenever one of them is changed.

Therefore a developer should make sure that any trait on a `BaseDataSourceModel` subclass that is expected to cause a change in the format of the return values of the data source `slots` method has the `changes_slots=True` metadata.

The `ChangesSlotsModel` is an example subclass, which should only experience a `_trigger_changes_slots` method call when the `b` trait is changed

flongford: The `_trigger_changes_slots` method uses the fact that the `on_trait_change` decorator mutates the signature of its method to include objects corresponding to the updated trait: https://docs.enthought.com/traits/traits_user_manual/notification.html#dynamic-notification

    handler(object, name, old, new)

force_bdss/data_sources/data_source_utilities.py
```diff
@@ -0,0 +1,196 @@
...
+class TraitSimilarityError(Exception):
+    """Reports the failure of a trait_similarity_check."""
+
+
+def merge_trait(source, target, name):
+    """Performs a merge of trait `name` between source and target
+    HasTrait objects. This is achieved by assigning the source
+    attribute onto the target if it has a non-default value.
+    Otherwise, the target attribute is assigned onto the
+    source.
```
Corwinpro: What is the use case of assigning the target attribute onto the source?

force_bdss/data_sources/data_source_utilities.py
```diff
@@ -0,0 +1,196 @@
...
+                attr_name,
+                getattr(source, attr_name),
+                getattr(target, attr_name)
+            )
+        )
+        logger.exception(error_msg)
+        raise TraitSimilarityError(error_msg)
+
+    # Merge attribute `name` between source and target
+    merge_trait(source, target, name)
```
Corwinpro: Would it make sense to refactor this function to separate the `check` part from the `merge_trait` part? Right now it seems that we are doing two things at a time. Maybe it could only perform the `if failed_attr: ...` verification, and return nothing if everything is fine. And then, in the method where `merge_trait_with_check` is called, we have two lines: one for the check, and the second for `merge_trait`.

force_bdss/data_sources/data_source_utilities.py
Corwinpro: See new PR

Corwinpro:
The `merge_trait_with_check` does two things: verifies there are no `failed_attr`ibutes, and then `merge_trait` for all `attributes`. And, as far as I can see, it is only used in `base_data_source_model._assign_slot_info`. 

My feeling is that we are trying to do two things at a time. I would consider this instead:
- `merge_trait_with_check`: remove the `merge_trait` functionality and rename to indicate this 
- `_assign_slot_info`: add `merge_trait` after the `merge_trait_with_check` which now only performs the check

Can we refactor it this way?

Corwinpro:
The change of `type` in the eggbox example doesn't seem to update the slots type.

flongford:
> The change of `type` in the eggbox example doesn't seem to update the slots type.

I don't think it will do in the UI until we're using the functionalities introduced in #230

flongford:
> The `merge_trait_with_check` does two things: verifies there are no `failed_attr`ibutes, and then `merge_trait` for all `attributes`. And, as far as I can see, it is only used in `base_data_source_model._assign_slot_info`.
> 
> My feeling is that we are trying to do two things at a time. I would consider this instead:
> 
> * `merge_trait_with_check`: remove the `merge_trait` functionality and rename to indicate this
> * `_assign_slot_info`: add `merge_trait` after the `merge_trait_with_check` which now only performs the check
> 
> Can we refactor it this way?

Good idea - I have attempted to implement these changes in the latest commits, please let me know if they seem appropriate

Issue #240, opened 2019-12-05
Adding generic MCOParameter subtypes


Corwinpro:
Currently example plugins are responsible for creation of custom `MCOParameter` types, relevant to their need. At this stage, we can start thinking about adding new generic `MCOParameter`s types. Essentially, these are already implemented in ITWM example, and some other examples. We would benefit from reusing those, and it would be easier to maintain them as a part of the BDSS core.

I would suggest these types:

```
# constant-valued dummy parameter
class FixedMCOParameter(BaseMCOParameter): pass

# numerical parameter with discrete values
class ListedMCOParameter(BaseMCOParameter): pass

# numerical parameter with continuous values in some range
class RangedMCOParameter(BaseMCOParameter): pass

# categorical unordered parameters
class CategoricalMCOParameter(BaseMCOParameter): pass
```

We can also think about vector-values parameters, if needed.

Corwinpro:
Note: there should be an obvious API to communicate the type of the `MCOParameter` to external data types, e.g. `nevergrad`.

Corwinpro:
Proposed `MCOParameters` are implemented in #242

PR #242, opened 2019-12-06
Implementing generic MCOparameter types

Corwinpro:
This PR approaches #240 

We introduce extension to `BaseMCOParameter`: 4 more specific (though still generic) types for `MCOParameter`s:

- `FixedMCOParameter`: constant-valued dummy parameter
- `RangedMCOParameter`: Numerical MCO parameter that continuously ranges between lower and upper floating point values
- `ListedMCOParameter`: Listed MCO parameter that has discrete numerical values
- `CategoricalMCOParameter`: implements unordered, discrete valued, categorical data. Available categorical values are strings

The purpose of the PR is to move these objects from separate example plugins into core, and maintain them as a part of `force-bdss`.

We also include necessary tests.

### DIFF
--- a/ci/__main__.py
+++ b/ci/__main__.py
@@ -27,6 +27,7 @@ ADDITIONAL_CORE_DEPS = [
     "numpy==1.15.4-2"
 ]
 
+
 @click.group()
 def cli():
     pass
--- a/ci/__main__.py
+++ b/ci/__main__.py
@@ -23,6 +23,9 @@ PIP_DEPS = [
     "stevedore==1.30.1"
 ]
 
+ADDITIONAL_CORE_DEPS = [
+    "numpy==1.15.4-2"
+]
 
 @click.group()
 def cli():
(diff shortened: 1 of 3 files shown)


force_bdss/mco/parameters/mco_parameters.py
```diff
@@ -0,0 +1,147 @@
...
+    upper_bound = Float(100.0)
+
+    n_samples = Int(5)
+
+    sample_values = Property(
+        depends_on="lower_bound,upper_bound,n_samples", visible=False
+    )
+
+    def _get_sample_values(self):
+        return np.linspace(self.lower_bound, self.upper_bound, self.n_samples)
```
flongford: Can we return this as a list? It's inefficient but at the moment the BaseMCO classes are all working with interfaces involving basic python types.

Corwinpro: Thank you, this will be `list(np.linspace(...))` instead.

ci/__main__.py
```diff
@@ -23,6 +23,9 @@
     "stevedore==1.30.1"
 ]
 
+ADDITIONAL_CORE_DEPS = [
+    "numpy==1.15.4-2"
```
flongford: Good, that's the same version as WfManager

Issue #249, opened 2020-01-07
Standard IO MCOCommunicator refactoring


Corwinpro:
Currently multiple plugins implement identical `MCOCommunicator` class which handles standard IO communication protocol (see for example [this](https://github.com/force-h2020/force-bdss-plugin-itwm-example/blob/master/itwm_example/mco/mco_communicator.py)

Developers and users would benefit from refactoring this communicator class into the bdss core application. This will save us time on maintaining the identical code across multiple plugins.

PR #250, opened 2020-01-07
WIP: DEV: Standard IO MCOCommunicator

Corwinpro:
This PR approaches #249 

We implement the `stdin`/`stdout` MCOCommunicator to the core BDSS.

The ITWM and other plugins now can import this implementation instead of copying the `MCOCommunicator` code.


Corwinpro:
This `MCOCommunicator` class may be in fact redundant.

Issue #254, opened 2020-01-08
Creating ListedMCOParameter in UI


Corwinpro:
When we attempt to create a new `ListedMCOParameter` using the workflow manager UI, an error is raised:
```
AttributeError: 'ListedMCOParameter' object has no attribute 'value'
```

The problem appears to be that the `value` attribute of the `levels` trait does not exist. Potential cause of the bug is a weird behaviour of the `List(Either(Int, Float), ...)`.

Corwinpro:
Solved in #255

PR #255, opened 2020-01-08
BUG: creating ListedMCOParameter via UI

Corwinpro:
This PR approaches #254 

We fix the bug that appears in workflow manager, when a new `ListedMCOParameter` is created. Apparently, there is a undocumented behaviour of the `List(Either())` trait.

Now, instead of using `Either(Int, Float)` we use just `Float`. This does not affect any logic of the BDSS or plugins, and fixes the `ListedMCOParameter` instantiation via the workflow manager UI.

### DIFF
--- a/force_bdss/mco/parameters/mco_parameters.py
+++ b/force_bdss/mco/parameters/mco_parameters.py
@@ -1,6 +1,6 @@
 import numpy as np
 
-from traits.api import List, Unicode, Property, Float, ReadOnly, Int, Either
+from traits.api import List, Unicode, Property, Float, ReadOnly, Int
 from traitsui.api import ListEditor
 
 from .base_mco_parameter import BaseMCOParameter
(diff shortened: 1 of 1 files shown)


Corwinpro:
For our reference: this seems to be the bug.

```
from traits.api import List, HasTraits, Property, Float, Instance, Int, Either
from traitsui.api import View, Item, InstanceEditor


from traits.api import List, HasTraits, Property, Float, Instance, Int, Either
from traitsui.api import View, Item, InstanceEditor


class Parameter(HasTraits):
    levels = List(Either(Float, Int), value=[0.0])

    sample_values = Property(depends_on="levels", visible=False)

    def _get_sample_values(self):
        return sorted(self.levels)


class ParameterView(HasTraits):
    parameter = Instance(Parameter)

    def default_traits_view(self):
        return View(Item("parameter", editor=InstanceEditor(), style="custom"))


view = ParameterView(parameter = Parameter())
view.configure_traits()
```

Issue #246, opened 2020-01-06
Workflow has a misleading `mco` attribute


Corwinpro:
Class `Workflow` has a `mco = Instance(BaseMCOModel, allow_none=True)` attribute.

It seems to be very confusing that it is called `mco` instead of `mco_model`, especially considering that there is a separate `BaseMCO` entity.

I can suggest changing the attribute name to `mco_model` - this better reflects the actual object type corresponding to this `Workflow` attribute.

flongford:
This seems like a very sensible idea to me

Corwinpro:
FIxed in #257

PR #257, opened 2020-01-08
Renamed the Workflow.mco to Workflow.mco_model

Corwinpro:
This PR approaches #246 

### Summary

The attribute `mco` of the `Workflow` class refers to the `Instance(BaseMCOModel)`. There are such entities as `BaseMCO` alongside the `BaseMCOModel`, and therefore this naming is misleading. We approach to resolve this inconsistency.

### Done

- We rename the attribute to `Workflow.mco_model`, and change all implicit and explicit references to `mco` to `mco_model`.

### Compatibility

This change doesn't breaks the backwards compatibility for workflow json files but throws a deprecation warning. We suggest that all workflow json file `workflow["mco"]` elements must be renamed to `workflow["mco_model"]`. In addition, the `force-wfmanager` requests to `Workflow.mco` should be changed to `Workflow.mco_model`.

### DIFF
--- a/force_bdss/app/workflow_evaluator.py
+++ b/force_bdss/app/workflow_evaluator.py
@@ -1,8 +1,11 @@
 import logging
 
 from traits.api import (
-    HasStrictTraits, Instance, Unicode, provides,
-    DelegatesTo
+    HasStrictTraits,
+    Instance,
+    Unicode,
+    provides,
+    DelegatesTo,
 )
 
 from force_bdss.core.data_value import DataValue
@@ -22,7 +25,7 @@ class WorkflowEvaluator(HasStrictTraits):
     workflow = Instance(Workflow)
 
     #: A reference to the mco model of the workflow instance.
-    mco_model = DelegatesTo('workflow', prefix='mco')
+    mco_model = DelegatesTo("workflow", prefix="mco_model")
 
     #: The path to the workflow file.
     workflow_filepath = Unicode()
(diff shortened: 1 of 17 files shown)


flongford:
I think we will need to create a PR for the `force_wfmanager` too - when adding a new MCO in the WfManager, a message appears: 

    traits.trait_errors.TraitError: The 'mco' trait of a Workflow instance is 'read only'.

Ironically, this looks like the failing line:

    File "/Users/flongford/FORCE/force-wfmanager/force_wfmanager/ui/setup/workflow_view.py", line 122, in set_mco
        self.model.mco = mco_model

force_bdss/app/workflow_evaluator.py
```diff
@@ -22,7 +25,7 @@ class WorkflowEvaluator(HasStrictTraits):
     workflow = Instance(Workflow)
 
     #: A reference to the mco model of the workflow instance.
-    mco_model = DelegatesTo('workflow', prefix='mco')
+    mco_model = DelegatesTo("workflow", prefix="mco_model")
```
flongford:
```suggestion
    mco_model = DelegatesTo("workflow", prefix="_mco_model")
```

force_bdss/core/workflow.py
```diff
@@ -12,11 +12,31 @@
...
+    #: Can be None if no MCOModel has been specified yet.
+    _mco_model = Instance(
+        BaseMCOModel, allow_none=True  # , transient=True, visible=False
+    )
+
+    #: Deprecated MCO attribute references to the mco_model attribute.
+    #: It is present to prevent possible dependencies API breaks.
+    mco = Property(depends_on="mco_model")  # , transient=True, visible=False)
+
+    mco_model = Property(depends_on="mco_model")
```
flongford:
```suggestion
    mco_model = Property(depends_on="_mco_model")
```

Corwinpro: Brilliant. Thank you.

PR #263, opened 2020-01-15
MAINT: Workflow IO classes structure and error handling refactoring

Corwinpro:
This PR approaches general refactoring of the BDSS `WorkflowReader` class and its integration with `WorkflowFile`.

### Summary

Most of the changes introduced in this PR were implemented in #266. Please find the description of the new features there.

We introduce minor refactoring to the structure of the `WorkflowReader` class. We move the method that reads the json file from the `WorkflowFile.read` to the `WorkflowReader.load` method, such that the workflow json file is opened and closed before any further data processing happens.

### Done

- Workflow file version check is refactored
- Top level `read_dict` method is refactored, and can be called with a dictionary as an argument
- Tests for `WorkflowReader` and `WorkflowWriter` updated to reflect the updated `WorkflowReader.read` method signature
- Added two new test fixtures, which were previously hardcoded into the test suites. 
- Added missing tests in `WorkflowWriter` (no code changes)


force_bdss/io/workflow_writer.py
```diff
@@ -1,12 +1,12 @@
...
-from traits.api import HasStrictTraits, Unicode
+from traits.api import HasStrictTraits, Str
 
 
 class WorkflowWriter(HasStrictTraits):
     """A Writer for writing the Workflow onto disk.
     """
 
-    version = Unicode('1')
+    version = Str("1")
```
flongford: Nice, this is something we can look out for in upcoming PRs - I will make an issue to remind us

Corwinpro: Thank you, I guess the `Unicode` trait was needed to handle `python2.*` versions in the first place. Which is not supported anymore.

flongford:
This is looking great, many thanks

However, loading in a file with `force_wfmanager` results in a `TypeError` - we will need to create a separate PR in `force-wfmanager` repo for this fix and release both at the same time.

flongford:
As discussed offline, we could benefit from addressing #264 at the same time

Issue #264, opened 2020-01-16
Workflow object can implement (de)serialization


Corwinpro:
[`Workflow` class](https://github.com/force-h2020/force-bdss/blob/f3743081d049eb7d1874e05b06b96ba4110ec7a0/force_bdss/core/workflow.py#L26) doesn't implement (de)serialization methods of itself into a json-compatible format, as well as can't create a complete instance of itself given a nested dictionary.

Currently, the `WorkflowReader` and `WorkflowWriter` classes are responsible for doing that. They seem to be separated from the `Workflow` class, despite that their area of responsibility is only to parse the `Workflow`-related data files.

We can consider refactoring this unnecessary separation of responsibilities, and move the `read` / `write` methods from `WorkflowReader` and `WorkflowWriter` to `Workflow`.

Corwinpro:
@flongford What do you think about this? Does this issue make sense?

flongford:
> @flongford What do you think about this? Does this issue make sense?

This makes sense - after chatting offline, we decided to implement a `__get_state__` method to the `Workflow` class that extracts all information of an instance into a dictionary containing basic python objects. 

This information should be able to be passed directly into the `__init__` method of the same class to create another instance possessing equivalent information.

The `WorkflowWriter` and `WorkflowReader` classes are then only used to save or load file objects in a certain format, rather than determining how the state of a `Workflow` instance is represented.

Corwinpro:
FYI: every `ContributedUI` class must ensure there is a version key in the `workflow_data` attribute. This will indicate the workflow data json version.

PR #266, opened 2020-01-17
DEV: Workflow serialization

Corwinpro:
This PR approaches #264 

### Summary

We work towards a standard way to serialize objects that are part of the `Workflow`. As usual, we use the `__getstate__` method for that. The `WorkflowWriter` now can directly access the `workflow.__getstate__()` dictionary instead of explicitly serializing the workflow data. The `WorkflowReader` can read and preprocess workflow data from a file, and create a `Workflow` instance. 

`Workflow` now implements a class method `.from_json` that creates a `Workflow` instance from a raw dict with necessary information.

The workflow json file version 1.1 now uses the following format for execution layer: the execution layers now explicitly contain the `"data_sources"` key. If an `ExecutionLayer` would require anything else in addition to `data_sources`, that would be easy to (de)serialize.

```
"execution_layers": [
    {
        "data_sources": [
            {
                "id": "force.bdss.enthought.plugin.test.v0.factory.dummy_data_source",
                "model_data": {
                    "input_slot_info": [{"name": "input_slot_name"}],
                    "output_slot_info": [{"name": "output_slot_name"}]
                }
            }
        ]
    }
]
```

### Done

- Changed the `BaseModel.__getstate__` signature: now it returns a `dict` with an `"id" : factory_id` entry and a `"model_data" : data` entry, instead of just `data`. 
- The `WorkflowWriter._extract_*` methods are removed as redundant. The `workflow.__getstate__` method now handles all the internal serialization.
- The `WorkflowReader` class now supports a new version of the workflow file format, `"1.1."`. The `_extract_execution_layers` method has been changed - now it verifies which file format is given, and parses the json data accordingly.
- Many test updates, including removing redundant tests.
- Test fixture for `WorkflowReader` / `WorkflowWriter` tests is updated to version 1.1.
- Basic deseralization for `Workflow` and `Workflow`'s attributes, via `klass.from_json(factory, json_data)` method.
- Removed `Workflow`-generating methods from `WorkflowReader` class. Now workflow instantiation errors should be raised directly by the `from_json` methods of the workflow attributes.
- `WorkflowReader.parse_data(data: Dict)` is a new class method of the workflow reader, that allows us to process the `json_data` of a workflow, without instantiating a `WorkflowReader` object. Once a workflow file has been read, we can parse the data from the file without reading the file again.


force_bdss/io/tests/test_workflow_reader.py
```diff
@@ -31,61 +36,103 @@ def setUp(self):
...
-    def test_initialization(self):
-        self.assertEqual(self.wfreader.factory_registry, self.registry)
-
-        workflow = self.wfreader.read(self.working_data)
-
-        self.assertIsInstance(workflow, Workflow)
-
     def test_load_data(self):
         data = self.wfreader.load_data(self.working_data)
         control_dict = {
```
flongford: Is it possible to add an additional dictionary to test continuing support for version 1 workflow JSONs?

force_bdss/io/tests/test_workflow_writer.py
```diff
@@ -25,9 +24,28 @@ def setUp(self):
...
+                    self.data_source_factory.create_model(),
+                    self.data_source_factory.create_model(),
+                ]
+            ),
+            ExecutionLayer(
+                data_sources=[self.data_source_factory.create_model()]
+            ),
+        ]
+        return wf
+
```
flongford: I believe we currently don't have a `ProbeWorkflow` class or equivalent - this useful code could be refactored later into a factory for creating such an object

force_bdss/io/workflow_writer.py
```diff
@@ -6,7 +6,7 @@ class WorkflowWriter(HasStrictTraits):
     """A Writer for writing the Workflow onto disk.
     """
 
-    version = Str("1")
+    version = Str("1.1")
```
flongford: Maybe make this `ReadOnly("1.1")`?

force_bdss/io/workflow_writer.py
```diff
@@ -22,72 +22,14 @@ def write(self, workflow, path, *, mode="w"):
...
-        """
-        Extracts the data from an external model and returns its dictionary
-        """
-        state = model.__getstate__()
-
-        return {"id": model.factory.id, "model_data": state}
+        return workflow.__getstate__()
 
 
 def pop_recursive(dictionary, remove_key):
```
flongford: It feels like `pop_recursive`, `pop_dunder_recursive` and `nested_getstate` could be moved to the same module as `BaseModel` since they are now used at a more fundamental level

force_bdss/io/tests/test_workflow_reader.py
Corwinpro: Sorry, it was in my "to push" stack. Should be there now.

force_bdss/io/workflow_writer.py
Corwinpro: Sounds good. I need to have a look at that, since it might be that too many plugins import these functions from here. I will create an issue for this instead, if you don't mind.

Corwinpro: Done.

force_bdss/io/tests/test_workflow_reader.py
Corwinpro: Tests added.

force_bdss/io/workflow_reader.py
```diff
@@ -80,36 +64,21 @@ def read(self, path):
...
         """
         json_data = self.load_data(path)
 
-        _ = self._extract_version(json_data)
-        workflow = self._extract_workflow(json_data)
+        self.workflow_format_version = self._extract_version(json_data)
+
+        workflow_data = self._preprocess_workflow_data(json_data)
+
+        workflow = Workflow.from_json(self.factory_registry, workflow_data)
```
flongford: Is it possible to create a `parse_data` pubic method to expose this section without needing to read the file?

It would be handy to use this in `force_wfmanager` (and would also make the mocking / unit testing a lot simpler!), where we can load in project files without having to read the same file twice.

Corwinpro: I quite agree, that would be very useful. I am working on this now. Thank you!

Corwinpro: Thank you, now we can use `WorkflowReader.parse_data(json_data)`.

PR #268, opened 2020-01-23
MAINT: remove workflow evaluator

### LINKED ISSUE #267: WorkflowEvaluator adds unnecessary abstraction layer
We can notice that the `WorkflowEvaluator` class is used only in the `OptimizeOperation.run` method. The `WorkflowEvaluator` functionality adds very little to the structure and coherence of the workflow execution, so it might be that this extra layer between the `Workflow` objects and the `OptimizeOperation` is not justified.

It might be beneficial to remove this class, unless there is a clear evidence that the `Workflow` itself can't implement the `evaluate` method.

I can suggest changing the `Workflow` signature to something like

```
@provides(IEvaluator)
class Workflow(HasStrictTraits):
    ...
    def evaluate(self, parameter_values):
        ...
```
This would do the same job as the `WorkflowEvaluator` in a more explicit way.

Corwinpro:
This PR closes #267 

### Summary

We remove the `WorkflowEvaluator` class and delegate the `evaluate` method directly to the `Workflow`.

Currently, the `WorkflowEvaluator` is a redundant layer between the `Workflow` instance and the `OptimizeOperation.run` method. This PR resolves this indirect dependency.


Corwinpro:
Note: the `SubprocessWorkflowEvaluator` class in `force-bdss-plugin-enthought-example/enthought_example/example_evaluator/subprocess_workflow_evaluator.py` requires `WorkflwoEvaluator`. We would need to fix that if this PR is merged.

Corwinpro:
The `Workflow` class can benefit from either an additional method `_subprocess_evaluate` or a `SubprocessEvaluator` mixin, that provide a functionality to execute the workflow in a subprocess. As an example, [this class](https://github.com/force-h2020/force-bdss-plugin-enthought-example/blob/aebb6dbc7d2dd282292e7a859ea23a68611f43de/enthought_example/example_evaluator/subprocess_workflow_evaluator.py#L12) can be used as a reference.

PR #273, opened 2020-01-28
MAINT: using tempfile instead of persistent `tmp.json` in tests

Corwinpro:
### Summary

This small PR uses `tempfile` module to create temporary workflow json files for tests, instead of a creating an annoyingly persistent `tmp.json` in the project folder.

### Done

- Use `tempfile` in tests and removed `tmp.json` from project folder.

### DIFF
--- a/force_bdss/io/tests/test_workflow_writer.py
+++ b/force_bdss/io/tests/test_workflow_writer.py
@@ -1,5 +1,6 @@
 import json
 import unittest
+import tempfile
 
 from force_bdss.core.execution_layer import ExecutionLayer
 from force_bdss.core.kpi_specification import KPISpecification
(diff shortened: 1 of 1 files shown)


flongford:
This is a very nice idea, thank you for suggesting this - I was also beginning to get annoyed with the pesky `temp.json` files!

However, unfortunately I think the `tempfile.mktemp` method is now deprecated (https://docs.python.org/3.6/library/tempfile.html#tempfile.mktemp). Could we implement a similar procedure using `tempfile.NamedTemporaryFile` instead?

Maybe also use a context manager, or include `os.unlink(filename)` at the end of the test suite to make sure it gets cleaned up afterwards.

Corwinpro:
Very nice idea considering it was me who put the `tmp.json` there. Fixing the deprecated issue now.

flongford:
> Very nice idea considering it was me who put the `tmp.json` there. Fixing the deprecated issue now.

Ha ha ha, you created a monster

PR #274, opened 2020-01-28
ENH: Include WeightedMCOStartEvent and WeightedMCOProgressEvent

flongford:
This PR includes two new subclasses:
- `WeightedMCOStartEvent`, inheriting from `MCOStartEvent`
- `WeightedMCOProgressEvent`, inheriting from `MCOProgressEvent`

They are designed to be used in combination to report KPI weights generated by the `WeightedOptimizerEngine`. In order to do so, a `BaseMCOModel` will need to assign each class as `_start_event_type` and `_progress_event_type` attributes respectively.

It also refactors the deprecated `BaseMCO.notify_new_point` method to ensure backwards compatibility for the new way the `MCOStartEvent.serialize` routine now controls data reporting in https://github.com/force-h2020/force-wfmanager/pull/339

Consequently, the new default behaviour of the MCO will be not **NOT** report any weights data unless informed by an accompanying `WeightedMCOStartEvent`

Finally, we also remove the deprecated and unused `BaseMCOModel.notify_new_point` method


force_bdss/mco/base_mco.py
```diff
@@ -79,11 +79,12 @@ def notify_new_point(self, optimal_point, optimal_kpis, weights):
             A list of weight values from 0.0 to 1.0 that have been assigned
             for this point to each KPI.
         """
-        self.notify(MCOProgressEvent(
-            optimal_point=optimal_point,
-            optimal_kpis=optimal_kpis,
-            weights=weights,
-        ))
+        self.notify(
+            WeightedMCOProgressEvent(
```
Corwinpro: Can this be an `MCOProgressEvent` for simplicity? I understand this will be removed _soon_, but still.

force_bdss/core_driver_events.py
```diff
@@ -46,6 +46,19 @@ def serialize(self):
         return self.parameter_names + self.kpi_names
 
 
+class WeightedMCOStartEvent(MCOStartEvent):
+    """Initializes reporting of weights generated during an MCO by a
+    WeightedOptimizerEngine"""
+
+    def serialize(self):
```
Corwinpro: We don't use the `serialize` method of the parent class here, do we? Would you say this is because we want a certain format of the header, in particular `parameters` + `(kpi, kpi_weight)`?

force_bdss/core_driver_events.py
```diff
@@ -62,31 +75,37 @@ class MCOProgressEvent(BaseDriverEvent):
...
+
+    #: Weights assigned to each KPI during the MCO optimization
+    weights = List(Float())
+
+    def serialize(self):
+        """Overloaded method to provide weights alongside each
+        reported KPI"""
+        event_datavalues = [entry.value for entry in self.optimal_point]
+        for kpi, weight in zip(self.optimal_kpis, self.weights):
+            event_datavalues.extend([kpi.value, weight])
```
Corwinpro: (Same question as in the comment above)

force_bdss/core_driver_events.py
flongford: Yes, I believe in the original table display each weight was reported in a column directly to the right of the KPI, rather than being appended to the end.

I just overloaded the function rather than calling the super to put this for both `Weighted*` events

Corwinpro:
I guess the codecov is complaining about the `default_weights`. Would you like me to help you with that?

flongford:
> I guess the codecov is complaining about the `_weights_default `. Would you like me to help you with that?

Ah yes, if you are not busy then that would be great - I can come up with some quick tests otherwise

PR #276, opened 2020-01-30
CLN: Refactor modules and objects involved in BaseDriverEvent propagation

flongford:
### Summary

Introduces new `force_bdss.events` module to contain objects related to `BaseDriverEvent`s. Includes `EventNotifierMixin` class that contains `event` attribute and `notify` method to propagate event.

Also moves some generic code to `force_bdss.utilties` to avoid some import loops

### Change log
- New `force_bdss.events` and `force_bdss.utilities` modules
- Moved `pop_recursive`, `pop_dunder_recursive` and `nested_getstate` functions into `force_bdss.utilities`
- All `MCO*Event` objects contained in `force_bdss.events.mco_events` module
- New class: `EventNotifierMixin`
- Remove `force_bdss.core_driver_events` module


force_bdss/events/tests/test_base_driver_event.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+class TestBaseDriverEvent(unittest.TestCase):
+
+    def test_getstate_base_event(self):
+        event = BaseDriverEvent()
+        self.assertFalse(event.__getstate__())
+
+        event = DummyEvent(stateful_data=DataValue(value=1))
+        self.assertDictEqual(
```
Corwinpro: FYI the `__getstate__` format will change after the (de)serialization PR is merged. The event is now serialized to the same format as the `BaseModel`: `{"id": id, "model_data": data}`.

PR #278, opened 2020-01-31
FIX: BaseCSVWriter bug caused by WeightedMCOStartEvent serialize method mutating attribute

flongford:
### Summary

This PR fixes a bug caused by the `WeightedMCOStartEvent.serialize` method mutating the `parameters` attribute whilst creating a list of strings to report as a header. Consequently, the `BaseCSVWriter` did not report the correct number of columns for the format of data coming back from the BDSS in a `WeightedMCOProgressEvent`.

This also causes issues displaying results data in the WfManager, since the same events are used to report data through the ZMQ server.

Additionally, we move the `BaseCSVWriter*` classes to the `force_bdss.notification_listeners` module, since the all inherit from respective `BaseNotificationListener*` classes

### Change log

- `WeightedMCOStartEvent.serialize` includes `deepcopy` call so attributes are not mutated
- Unit test included in `test_base_csv_writer` module to ensure correct parsing of `WeightedMCOStartEvent`
- `force_bdss.io.base_csv_writer` module moved to `force_bdss.notification_listeners.base_csv_writer`


flongford:
> Would you mind if we add a bit more information on why do we move the BaseCSVWriter to the notification listeners?

All three `BaseCSVWriter` classes inherit from the `BaseNotificationListener` classes, so I would expect them to be in the same module

Corwinpro:
> Sure, all three `BaseCSVWriter` classes inherit from the `BaseNotificationListener` classes, so I would expect them to be in the same module. I can make this clearer in the description.

Thank you, you are absolutely right. Although now it feels that the `BaseCSVWriter` class should not be a subclass of `BaseNotificationListener`... which in fact is kind of weird.

PR #281, opened 2020-02-03
WIP: ENH: Integration of CUDS using osp-core package

flongford:
### NOTE: this PR requires a local installation of `osp-core==3.3.2`

Integration with CUDS-type ontologies is achieved using `osp-core` package. This PR expands upon the `CUBAType` attribute functionalities in order to allow for a simple python basic type and `TraitType` class lookup for a given CUBA key.

Consequently, we are now able to utilise powerful extended physical unit typing abilities provided by an ontology. We expect later extensions to expand upon this by integrating external software wrappers built around the `osp-core` ontology system.

### Change log:
- BDSS installation requires `osp-core==3.3.2` (expected to fail on Travis CI)
- New `Ontology` class, a subclass of `TraitType`, allowing for `HasTraits` subclasses to hold `OntologyNamespace` attributes
- New `IOntologyRegistry` class providing an interface between loaded `osp-core` ontologies and `CUBAType` objects
- New `CUBATypeMixin` class providing the ability to look up CUBA keys in a provided `IOntologyRegistry`
- `CUBAType` class now contains regex rules determined by `osp-core`


flongford:
This implementation is not ideal, since we end up needing a registry system for `CUBAType` objects. This system should ultimately come from the ontology modules themselves (we don't want to redesign global state).

In which case, we should redefine the `CUBAType` class as a traits wrapper around `OntologyAttribute`, which already has a reference to the ontology in its module path. However, this will break any existing plugin objects using simple for as `DataValue.type` objects.

flongford:
Note: the `CUBATypeMixin` class should also probably be inherited by `DataValue`

PR #282, opened 2020-02-04
WIP: FEAT: FileTreeBuilder class ported from force_gromacs

flongford:
This PR ports and extends the useful `FileTreeBuilder` class from `force_gromacs` repo.

It can be used to create a nested set of directories in a file system, multiple layers deep. An extra function is included in the `force_bdss.utilities` module to recursively search through a set of nested dictionaries and report back generated file paths.

### Example:

The file tree

```
root
├── folder_1
│   └── folder_1.1
└── folder_2
    ├── folder_2.1
    └── folder_2.2
    |   └── folder_2.2.1
    └── folder_2.3
```

Would be represented by the following `FileTreeBuilder` instance:

``` python
FileTreeBuilder(
       root='root',
       _file_tree={
           'folder_1': {
               'folder_1.1': {}
           },  
           'folder_2': {
              'folder_2.1': {},
              'folder_2.2': {
                  'folder_2.2.1': {}
              },
             'folder_2.3': {},
          }
     }
)
```
In general, the `_file_tree` attribute would not be passed as an argument, but can be constructed using the `FileTreeBuilder.add_path` and `FileTreeBuilder.add_folders` methods.

A possible set of calls to create this `_file_tree` attribute dictionary as shown above could be :

``` python
file_buildler = FileTreeBuilder(root='root')
file_builder.add_path('folder_1/folder_1.1')
file_builder.add_path('folder_2/folder_2.2/folder_2.2.1')
file_builder.add_folders('folder_2', ['folder_2.1', 'folder_2.3'])
```

After constructing the file tree, this can be built in the file system using

```python
file_builder.build_file_tree()
```

This is performed safely - no existing folders or directories will be overwritten

### DIFF
--- a/force_bdss/io/file_tree_builder.py
+++ b/force_bdss/io/file_tree_builder.py
@@ -14,7 +14,7 @@ class FileTreeBuilder(HasStrictTraits):
 
     #: Location to create file tree in. (By default, the
     #: current working directory)
-    directory = Directory()
+    root = Directory()
 
     # --------------------
     #  Private Attributes
@@ -24,9 +24,7 @@ class FileTreeBuilder(HasStrictTraits):
     _file_tree = Dict(Directory, Dict)
 
     def __file_tree_default(self):
-        return collections.OrderedDict({
-            self.directory: {}
-        })
+        return collections.OrderedDict()
 
     # --------------------
     #   Private Methods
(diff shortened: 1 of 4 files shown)


flongford:
Closing since no longer required

Issue #283, opened 2020-02-05
BUG: nested_getstate removes Dict trait data


Corwinpro:
If an instance of `BaseModel` has a `Dict` attribute, this data will be ignored when parsing with `__getstate__`. I guess this happens inside the `nested_getstate`. 

Update:
This is a Traits behaviour. See the code below:
```
>>> from traits.api import HasTraits, Str, Int, Dict
>>> class foo(HasTraits):
...     string = Str("string")
...     integer = Int(42)
...     dictionary = Dict({"d": 1})
... 
>>> f = foo()
>>> f.__getstate__()
{'string': 'string', 'integer': 42, 'dictionary': {'d': 1}, '__traits_version__': '5.1.1'}
>>> f.string.__getstate__()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'str' object has no attribute '__getstate__'
>>> f.integer.__getstate__()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'int' object has no attribute '__getstate__'
>>> f.dictionary.__getstate__()
{'name': 'dictionary', 'name_items': 'dictionary_items'}
```
The difference is that the `Dict` is the only type that implements `__getstate__`, and therefore doesn't raise an `AttributeError`.
To fix this, we could add this extra condition to the `nested_get_state`:
```
            if isinstance(state_dict[key], dict):
                continue
```
to the `nested_getstate` function.

Corwinpro:
Closed by #284

PR #284, opened 2020-02-05
ENH: added explicit dict type case for nested getstate

Corwinpro:
This PR approaches #283 

### Summary

We add explicit treatment of `dict` type parsing in the `nested_getstate` function. Now, instead of using the built-it Traits `__getstate__` method of `Dict` (which doesn't return the `Dict`'s data but rather the traits representation of that `Dict`), we retain the top-level Traits `__getstate__`.

### Done

- Added extra case to `nested_getstate`:
```
 elif isinstance(state_dict[key], dict):
     continue
```
- Added simple tests to verify that the `Dict` trait is parsed correctly.

### DIFF
--- a/force_bdss/utilities.py
+++ b/force_bdss/utilities.py
@@ -52,10 +52,7 @@ def nested_getstate(state_dict):
             if isinstance(state_dict[key], (tuple, list)):
                 state_dict[key] = [el.__getstate__() for el in state_dict[key]]
             elif isinstance(state_dict[key], dict):
-                for element in state_dict[key]:
-                    state_dict[key][element] = state_dict[key][
-                        element
-                    ].__getstate__()
+                continue
             else:
                 state_dict[key] = state_dict[key].__getstate__()
         except AttributeError:
--- a/force_bdss/utilities.py
+++ b/force_bdss/utilities.py
@@ -51,6 +51,11 @@ def nested_getstate(state_dict):
         try:
             if isinstance(state_dict[key], (tuple, list)):
                 state_dict[key] = [el.__getstate__() for el in state_dict[key]]
+            elif isinstance(state_dict[key], dict):
+                for element in state_dict[key]:
+                    state_dict[key][element] = state_dict[key][
+                        element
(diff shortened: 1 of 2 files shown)


force_bdss/utilities.py
```diff
@@ -51,6 +51,8 @@ def nested_getstate(state_dict):
         try:
             if isinstance(state_dict[key], (tuple, list)):
                 state_dict[key] = [el.__getstate__() for el in state_dict[key]]
+            elif isinstance(state_dict[key], dict):
```
flongford:
```suggestion
            elif not isinstance(state_dict[key], dict):
```
And remove lines 55 + 56?

Corwinpro: Thank you, this is a very good suggestions. To the best of my knowledge, these two approaches should be equivalent? If yes, I will try to think whether the old explicit case is better / worse from the readability point of view. I believe you are absolutely right that we can merge these three lines with one simple `not` statement; I am very bad however at correctly identifying the truthy statements with `not`.

I will look into this and come back to you asap, thank you.

flongford: No worries, yes, I think they should be equivalent. 

I sometimes get confused with nested logic statements, so reducing the number of lines can help simplify things. However, I also agree that using `not` is more implicit, so please use whatever way you think is more clear

Corwinpro: I have a feeling that it is not the only type we should treat explicitly. However, I can't narrow down the exact situation when this will break. Something like `d = Dict(value_trait = List(CustomHasTraits))` would definitely require a custom `__getstate__`.

After talking to Mark today, it seems that having a generic `__getstate__` is not the best idea overall. We should not hesitate to implement object-specific gestate methods, if we want to maintain different traits versions compatibility.

flongford: That is a very good point. I have been thinking along similar lines recently - or at least, feeling that we should probably leave the `__getstate__` super traits method alone and implement the serialization in a separate method.

PR #286, opened 2020-02-17
DEV: BDSS communication

Corwinpro:
This PR approaches [#352](https://github.com/force-h2020/force-wfmanager/issues/352)

### Summary

We introduce a `UIEventNotificationMixin` class that has two `threading.event` attributes to control - stop and pause - the BDSS execution. We add the stop and pause functionality to the `OptimizeOperation`: the `_deliver_event` method (which is called every time an event is fired by the `Workflow`) also reacts to the `_stop_event`, `_pause_event` changes via `OptimizeOperation .ui_event_response()` method.

If a `OptimizeOperation._pause_event` is unset, we wait for the UI to resume the run.
If a `OptimizeOperation._stop_event` is set, this indicates that the optimization should be terminated.

### Done
- Added a mixin class `UIEventNotificationMixin` that adds the `_stop_event`, `_pause_event` control API to Notification Listeners.
- Added `ui_event_response` method to the `OptimizeOperation`, that responds to the change of the threading stop and pause events. If the `_pause_event` is unset, we wait for the event to be set back. If the `_stop_event` is set, we do `sys.exit`
- The `OptimizeOperation. _initialize_listeners` method now tries to assign the  `_stop_event` and `_pause_event` to the listeners. These event will be only assigned to the listeners that are subclasses of `UIEventNotificationMixin`.

### To do
- Implement a better (safer) way to terminate the BDSS execution.


force_bdss/app/optimize_operation.py
```diff
@@ -114,6 +126,15 @@ def _deliver_event(self, event):
                 self._finalize_listener(listener)
                 self.listeners.remove(listener)
 
+        self.ui_event_response()
+
+    def ui_event_response(self):
```
flongford: Docstring here would be great

force_bdss/app/optimize_operation.py
```diff
@@ -39,6 +41,16 @@ class OptimizeOperation(HasStrictTraits):
     #: The notification listener instances.
     listeners = List(Instance(BaseNotificationListener))
 
+    _stop_event = Instance(ThreadingEvent, visible=False, transient=True)
+
+    _pause_event = Instance(ThreadingEvent, visible=False, transient=True)
+
```
flongford: @Corwinpro Docstrings for these traits would be great too cheers

force_bdss/ui_hooks/ui_notification_mixins.py
```diff
@@ -0,0 +1,29 @@
+from threading import Event as ThreadingEvent
+
+from traits.api import Instance, HasStrictTraits
+
+
+class UIEventNotificationMixin(HasStrictTraits):
```
flongford: Docstring explaining when you would use this class would be great too

force_bdss/ui_hooks/ui_notification_mixins.py
```diff
@@ -0,0 +1,33 @@
...
+
+    def set_stop_event(self, event):
+        self._stop_event = event
+
+    def set_pause_event(self, event):
+        self._pause_event = event
+
+    def send_stop(self):
+        self._pause_event.set()
+        self._stop_event.set()
```
flongford: Note: we should set the stop first before unpausing (otherwise the run may continue depending on when the stop event comes in)

PR #287, opened 2020-02-19
MAINT: Workflow backwards compatiblity

### LINKED ISSUE #271: Backwards compatibility of WorkflowReader
Could we reintroduce the backwards compatibility of reading the `mco` key instead of `mco_model` into the `WorkflowReader` for JSONs using version 1 formatting?

It would also therefore be great to document the differences between version 1 and 1.1

Corwinpro:
This PR closes #271 

### Summary

We add a `"version"` specific rule in the `WorkflowReader._preprocess_workflow_data()` method, that expect the `"mco"` keyword in the workflow configuration file of `"version": "1"`.  

The workflow config files of version `"1"` use `"mco"` as the key to identify the mco model parameters, so we restore the ability to read old configuration file with `"mco"` instead of `"mco_model"`.


flongford:
Note: This branch was successfully merged to master, though something happened to GitHub at the time and so the PR request was not updated successfully

PR #290, opened 2020-02-21
DEV: DataSource events

Corwinpro:
### Summary

This PR implements new `DataSourceEvents` that are fired before and after each `DataSource` is executed. This allows more precise control of the Workflow execution, e.g., when stop / pause events are fired by the WorkflowManager.

### Fixes

- We fix the undesired behaviour described in [#361](https://github.com/force-h2020/force-wfmanager/issues/361) by introducing a new event subtype `UIEventMixin`.

### Done

- Implemented `DataSourceStartEvent`, `DataSourceFinishEvent` subclasses of the `BaseDriverEvent`. Currently, no specific functionality is added to these events.
- Implemented private `_run` method that wraps around the standard public `BaseDataSource.run` method, and notifies the start and finish of the `DataSource` execution.
- `ExecutionLayer` now uses the `DataSource._run` method instead of the `DataSource .run` to trigger the start and finish events on the `DataSource` instance.


Corwinpro:
> So far this is looking great, and works very well, thank you.
> 
> I only have one comment worth investigating - testing with the ITWM example, I can see that the calculation time of each optimal point is slower than using the `master` branch of `force_bdss`. Do you find the same? We might want to do some benchmarking just to check.

Thank you, I am investigating this. Apparently the WfManager is receiving too many messages from the BDSS.

PR #295, opened 2020-02-25
DEV: Vector MCO parameter

Corwinpro:
This PR approaches [#11](https://github.com/force-h2020/force-bdss-plugin-nevergrad/issues/11)

### Summary

We introduce a `RangedVectorMCOParameter` which is an extension to the existing `RangedMCOParameter`, but implements vector-valued bounds and values.

### Done

- Refactored the `View` of the `RangedMCOParameter` to add `RangeEditor` view to the `initial_value`.
- Implemented the `RangedVectorMCOParameter`. The dimensions of the vector-valued attributes are defined by the `dimension` attribute. The `lower_bound, upper_bound, initial_value` lists add / remove elements if the lengths are different from the `dimension` value. The bounds view is a standard `ListEditor` enhanced with a `style="readonly"` option.
- Implemented `verify` methods for the `RangedMCOParameter`s: we test whether the upper bound is above the lower bound, and if the initial value is within these bounds. 
- Also includes check to ensure each vector property has the expected dimensionality, since this could be incorrectly assigned in a workflow JSON.
- Added `verify=True` to the `name`, `type` traits of the `BaseMCOParameter`
- Added `verify=True` to the bounds and the initial value of the `RangedMCOParameter`s.

### DIFF
--- a/force_bdss/mco/parameters/tests/test_mco_parameters.py
+++ b/force_bdss/mco/parameters/tests/test_mco_parameters.py
@@ -100,6 +100,34 @@ class TestRangedMCOParameter(TestCase):
         self.assertEqual(5, parameter.n_samples)
         self.assertEqual(100.0, parameter.initial_value)
 
+    def test_verify(self):
+        parameter = RangedMCOParameter(
+            self.factory,
+            lower_bound=10,
+            upper_bound=20,
+            initial_value=30,
+            name="name",
+            type="type",
+        )
+        errors = parameter.verify()
+        messages = [error.local_error for error in errors]
+        self.assertEqual(1, len(messages))
+        expected_message = (
+            "Initial value of the Ranged parameter must be "
+            "withing the lower and the upper bounds."
+        )
+        self.assertEqual(expected_message, messages[0])
+
+        parameter.lower_bound = 25
+        errors = parameter.verify()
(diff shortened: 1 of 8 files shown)


setup.py
```diff
@@ -1,7 +1,7 @@
 import os
 from setuptools import setup, find_packages
 
-VERSION = "0.4.0"
+VERSION = "0.5.0dev"
```
flongford: Should be taken care of by #296

force_bdss/mco/parameters/tests/test_mco_parameters.py
```diff
@@ -98,6 +100,173 @@ def test_custom_ranged_parameter(self):
...
+        )
+        self.assertEqual(expected_message, messages[0])
+
+        parameter.initial_value = 0
+        errors = parameter.verify()
+        messages = [error.local_error for error in errors]
+        self.assertEqual(1, len(messages))
+        expected_message = (
+            "Initial value of the Ranged parameter must be "
+            "withing the lower and the upper bounds."
```
flongford: Docstring typo: "withing"

Corwinpro: pengwings

PR #297, opened 2020-02-26
ENH: MCO runtime event

Corwinpro:
### Summary

This PR attempts to build a foundation for the `BaseDriverEvent`s that are fired during the `Workflow` execution. We introduce a `MCORuntimeEvent` type. The `DataSourceEvent`s are now subclasses of the `MCORuntimeEvent`.

### Done

- Added `MCORuntimeEvent` to the MCO events. 
- `DataSourceStartEvent ` and `DataSourceFinishEvent ` events are now subclasses of `MCORuntimeEvent`

### DIFF
--- a/force_bdss/events/mco_events.py
+++ b/force_bdss/events/mco_events.py
@@ -96,6 +96,9 @@ class MCORuntimeEvent(BaseDriverEvent):
      the Workflow or the UI, this event would be an MCORuntimeEvent.
     """
 
+    #: An identification label (name) of the MCORuntimeEvent
+    name = Str(allow_none=False)
+
 
 class WeightedMCOStartEvent(MCOStartEvent):
     """Initializes reporting of weights generated during an MCO by a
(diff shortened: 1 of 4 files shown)


force_bdss/events/mco_events.py
```diff
@@ -90,6 +90,16 @@ def from_json(cls, json_data):
         return cls(**data)
 
 
+class MCORuntimeEvent(BaseDriverEvent):
+    """ The base class for the MCO events fired during the workflow
+     execution. For example, if a DataSource is designed to notify
+     the Workflow or the UI, this event would be an MCORuntimeEvent.
```
flongford: Can we make it clear why this class does not inherit from `UIEventMixin` and therefore will not be reported back to the `WfManager` by default? Otherwise the docstring here might mislead a developer

Corwinpro: That is true, thank you. I will find a better explanation for this case.

PR #305, opened 2020-03-23
serialize returns pass

### LINKED ISSUE #299: UIEventMixin must be abstract base class
The only responsibility of the `UIEventMixin` is to provide a way to communicate with the WorkflowManager. The `MCOStart`, `Progress`, and `Finish` events already implement serialization method `.serialize()`.

Now, as we develop a generic way for the WorkflowManager to receive events, it is apparent that all  `UIEventMixin` subclasses must implement a `.serialize` method.  We can consider adding an abstract `serialize` method to the `UIEventMixin` to enforce this requirement.

sparsonslab:
MCOFinishEvent.serialize returns pass. Closes #304 and closes #299

### DIFF
--- a/force_bdss/events/mco_events.py
+++ b/force_bdss/events/mco_events.py
@@ -47,7 +47,7 @@ class MCOFinishEvent(BaseDriverEvent, UIEventMixin):
     """ The MCO driver should emit this event when the evaluation ends."""
 
     def serialize(self):
-        raise NotImplementedError
+        pass
 
 
 class MCOProgressEvent(BaseDriverEvent, UIEventMixin):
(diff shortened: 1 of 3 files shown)


force_bdss/events/mco_events.py
```diff
@@ -47,7 +47,7 @@ class MCOFinishEvent(BaseDriverEvent, UIEventMixin):
     """ The MCO driver should emit this event when the evaluation ends."""
 
     def serialize(self):
-        raise NotImplementedError
```
flongford: Could we put a docstring here: "We don't expect these events to carry any information that requires serialization"

sparsonslab: Can we do the coverage test locally before pushing the branch? Is there a python -m ci command?

flongford: `python -m ci coverage`

PR #313, opened 2020-03-25
CLN: Remove Deprecated code

### LINKED ISSUE #312: Remove Deprecated code
Some code were deprecated in version 4.0.0 and can be removed.

These include:
- The `BaseMCO.notify_progress_event` method. Plugin developers should use the new API introduced in version 4.0.0 and call `evaluator.mco_model.notify_progress_event` during `BaseMCO.run` instead.

- The `OptimizeOperation._deliver_event` should therefore no longer listen to `'mco:event'`, allowing us to remove the `OptimizeOperation.mco` attribute and refactor the run method not to include binding to a `BaseMCO` instance.

flongford:
The PR closes #312 

It removes both the `BaseMCO.notify_progress_event` method and the `OptimizeOperation.mco` attribute. Consequently, any existing `BaseMCO` subclasses will need to ensure they are no longer calling `self.notify_progress_event`, and instead replace this with `evaluator.mco_model.notify_progress_event` instead.

Related unit test code is also updated, and an extra test is included to ensure `UIEventNotificationMixin` classes have their pause and stop events set correctly during initialization.

# Change log:
- Removed `BaseMCO.notify` and `BaseMCO.notify_progress_event` methods
- Removed `BaseMCO.event` attribute
- Removed `OptimizeOperation.mco` attribute
- New probe class: `ProbeUIEventNotificationListener`

# Note
~We should probably check that all our existing `BaseMCO` classes in plugins are not still using `self.notify_progress_event` before merging~

### DIFF
--- a/force_bdss/app/optimize_operation.py
+++ b/force_bdss/app/optimize_operation.py
@@ -35,9 +35,6 @@ class OptimizeOperation(HasStrictTraits):
     #: The workflow instance.
     workflow = DelegatesTo("workflow_file")
 
-    #: The mco instance.
-    mco = Instance(BaseMCO)
-
     #: The notification listener instances.
     listeners = List(Instance(BaseNotificationListener))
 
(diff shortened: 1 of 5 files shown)


flongford:
Not sure what is going on with codecov - possibly this will be solved by merging #314

PR #314, opened 2020-03-26
Interface methods

### LINKED ISSUE #298: Some Interface subclasses contain abstract methods with implementations
There are some `Interface` subclasses that implement `pass` statements in their abstract methods, whereas a docstring would suffice.

- `IFactory`
- `IMCOFactory`
- `IUIHooksFactory`
- `IDataSourceFactory`
- `INotificationListenerFactory`

Additionally, by doing so we seem to be confusing the coverage, which counts these `pass` statements as missed lines. If we simply used docstrings, then this would not be an issue

sparsonslab:
Closes #298. What the functions do is fairly obvious from their names. More detailed docs are associated with the implementing base classes.

### DIFF
--- a/force_bdss/mco/i_mco_factory.py
+++ b/force_bdss/mco/i_mco_factory.py
@@ -26,22 +26,36 @@ class IMCOFactory(IFactory):
     )
 
     def get_model_class(self):
-        pass
+        """
+        :return: model class.
+        """
 
     def get_communicator_class(self):
-        pass
+        """
+        :return: model communicator class.
+        """
 
     def get_optimizer_class(self):
-        pass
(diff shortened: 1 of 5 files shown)


force_bdss/mco/i_mco_factory.py
```diff
@@ -26,22 +26,36 @@ class IMCOFactory(IFactory):
...
     def create_communicator(self):
-        pass
+        """
+        :return: communicator
+        """
 
     def parameter_factories(self):
-        pass
+        """
+        :return:
```
flongford: suggestion: `return mco parameter factories`?

PR #315, opened 2020-03-27
added start and finish events

### LINKED ISSUE #291: Create a public API to expand utilities of DataSourceStartEvent and DataSourceFinishEvent
Since #290, we have two new types of `BaseDriverEvent` that are used to improve the responsiveness of the BDSS a stop or pause event. They currently do not report any information back to the UI.

However, it is possible that a developer may desire to include extra functionalities to these events, such as passing data out to an external file or program. We could therefore allow for the contribution of custom start and finish events for a `BaseDataSourceModel`, in a similar way to the `BaseMCOModel` class

sparsonslab:
Closes #291. Added start and finish events, both with serialize methods, to base data source model. Just copies approach/code for mco model/events.

### DIFF
--- a/force_bdss/events/data_source_events.py
+++ b/force_bdss/events/data_source_events.py
@@ -14,14 +14,14 @@ class DataSourceStartEvent(MCORuntimeEvent):
     input_names = List(Str())
 
     def serialize(self):
-        """ Provides serialized form of MCOStartEvent for further data storage
+        """ Provides serialized form of DataSourceStartEvent for further data storage
         (e.g. in csv format) or processing.
 
 
         Usage example:
-        For a custom MCOStartEvent subclass, this method can be overloaded.
+        For a custom DataSourceStartEvent subclass, this method can be overloaded.
         An example of a custom `serialize` method would be:
-        >>> class CustomMCOStartEvent(DataSourceStartEvent):
+        >>> class CustomDataSourceStartEvent(DataSourceStartEvent):
         >>>
         >>>     def serialize(self):
         >>>         custom_data = [f"{name} data" for name in self.kpi_names]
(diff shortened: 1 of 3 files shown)


force_bdss/events/data_source_events.py
```diff
@@ -1,11 +1,62 @@
...
+        >>> class CustomMCOStartEvent(DataSourceStartEvent):
+        >>>
+        >>>     def serialize(self):
+        >>>         custom_data = [f"{name} data" for name in self.kpi_names]
+        >>>         return super().serialize() + custom_data
+        >>>
+
+        Returns:
+            List(Str): event parameters names and kpi names
+        """
```
flongford: Some MCO references in this docstring just need updating to DataSource.

The `self.kpi_names` reference can also be changed to `self.input_names`

flongford: Some MCO references in this docstring just need updating to DataSource.

The `self.kpi_names` reference can also be changed to `self.input_names`

force_bdss/events/data_source_events.py
```diff
@@ -1,11 +1,62 @@
...
```
flongford: Same as previous comment regarding docstring updates, though I guess `self.kpi_names` reference can be changed to `self.output_names`

force_bdss/events/tests/test_data_source_events.py
```diff
@@ -25,9 +25,19 @@ def test_getstate_finish_event(self):
...
+
+    def test_serialize_finish_event(self):
+        event = DataSourceFinishEvent()
+
+        self.assertListEqual(event.serialize(), [])
+
+    def test_serialize_start_event(self):
+        event = DataSourceStartEvent()
+
+        self.assertListEqual(event.serialize(), [])
```
flongford: Can we also provide tests to check the expected outcome should be when `event.input_names` or `event.output_names` are assigned values?

PR #316, opened 2020-03-27
Base operation threading

### LINKED ISSUE #310: Create BaseOperation class containing threading event handling
Since #286 we use the `threading` library to add pause and stop functionalities to the `OptimizeOperation` class.

These are yet to be implemented into the `EvaluateOperation` class, which should be interchangeable with `OptimizeOperation`, since it provides the `IOperation` interface. 

To achieve this we could create a `BaseOperation` class containing all generic features (including threading objects) shared by both `OptimizeOperation` and `EvaluateOperation`.

sparsonslab:
Closes #310. Create base class for EvaluateOperation and OptimizeOperation for common threading attributes.

### DIFF
--- a/force_bdss/app/base_operation.py
+++ b/force_bdss/app/base_operation.py
@@ -1,0 +1,48 @@
+from threading import Event as ThreadingEvent
+import logging
+
+from traits.api import (
+    DelegatesTo,
+    HasStrictTraits,
+    Instance,
+    provides
+)
+
+from .i_operation import IOperation
+from .workflow_file import WorkflowFile
+
+log = logging.getLogger(__name__)
+
+
+@provides(IOperation)
+class BaseOperation(HasStrictTraits):
+    """ Base class for EvaluateOperation and OptimizeOperation.
+    """
+
+    #: The workflow file being operated on.
+    workflow_file = Instance(WorkflowFile)
+
+    #: The workflow instance.
+    workflow = DelegatesTo("workflow_file")
+
+    #: Threading Event instance that indicates if the optimization operation
+    #: should be stopped.
+    _stop_event = Instance(ThreadingEvent, visible=False, transient=True)
+
+    #: Threading Event instance that indicates if the optimization operation
+    #: should be paused and then resumed.
+    _pause_event = Instance(ThreadingEvent, visible=False, transient=True)
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._stop_event = ThreadingEvent()
+        self._pause_event = ThreadingEvent()
+        self._pause_event.set()
+
+    def run(self):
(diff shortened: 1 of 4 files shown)


force_bdss/app/base_operation.py
```diff
@@ -0,0 +1,48 @@
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._stop_event = ThreadingEvent()
+        self._pause_event = ThreadingEvent()
+        self._pause_event.set()
+
+    def run(self):
+        """ Evaluate the workflow. """
+
+    def _deliver_event(self, event):
```
flongford: We can include the implementation of ``_deliver_event`` in both ``EvaluateOperation`` and ``OptimizeOperation`` here, since it will be shared between both classes.

By doing so it probably allows us to use the existing unit test code for this method to meet the Codecov threshold

force_bdss/app/base_operation.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._stop_event = ThreadingEvent()
+        self._pause_event = ThreadingEvent()
+        self._pause_event.set()
+
+    def run(self):
+        """ Evaluate the workflow. """
+
```
flongford: We should probably raise a `NotImplementedError` here for good measure

force_bdss/app/base_operation.py
sparsonslab: Optimizeoperation._deliver_event contains a lot more code, with the code that it shares with EvaluateOperation._deliver_event coming at the end. I didn't want to screw up the order of run, etc.

force_bdss/app/base_operation.py
sparsonslab: On another issue you suggested just putting a docstring in an empty base class implementation, rather than a NotImplementedError (I think maybe to do with unit testing).

sparsonslab: Or maybe that was because it was an interface? i.e.
interface = just docstring
base class = return error
Is that the pattern?

flongford: Yes, that's correct - an `Interface` class expects no implementation, whereas that is not the case for our base classes. If we wanted to we could turn `BaseOperation` into an `ABCHasTraits` class and use abstract methods here, but a more simple way is just to raise a `NotImplementedError` instead

force_bdss/app/base_operation.py
flongford: Ah, apologies - looking at this issue again I believe that the `listener` related code should also be ported to the base class. Only the `run` methods (and any related code) will end up being unique to both subclasses. Sorry, that wasn't made clear in the original issue

Issue #317, opened 2020-03-27
EvaluateOperation should verify Workflow first


flongford:
Currently, the `EvaluateOperation.run` method does not call the `Workflow.verify` method before evaluating the KPIs for a given set of parameter values. This is problematic, since it's possible that a single evaluation point could take a long time to execute, which would be wasted if it breaks at any point.

The `OptimiseOperation.run` method does, however, verify the `Workflow` object before performing its optimization task. Therefore we could also share the same code (possibly through a `BaseOperation` class #310) in order to resolve this.

sparsonslab:
I've tried porting `self.workflow_file.verify()` from the optimizer_operation, and the log.error code that follows it, into the base_operation in two ways:
1) putting it directly into `BaseOperation.run()` and then calling super in the run() of the optimizer and evaluator operations.
2) making a `BaseOperation.verify_workflow()` method that is called by run() of the optimizer and evaluator operations.
In both cases I get a heap of unit test errors, including ones in non-operation test (e.g. `test_cli_execution.py` that I can't untangle.
I might need some help on this one.

flongford:
> I've tried porting `self.workflow_file.verify()` from the optimizer_operation, and the log.error code that follows it, into the base_operation in two ways:
> 
> 1. putting it directly into `BaseOperation.run()` and then calling super in the run() of the optimizer and evaluator operations.
> 2. making a `BaseOperation.verify_workflow()` method that is called by run() of the optimizer and evaluator operations.

Just as a side comment, I would opt for option 2 here as our final approach, since it feels cleaner.

> In both cases I get a heap of unit test errors, including ones in non-operation test (e.g. `test_cli_execution.py`) that I can't untangle.
> I might need some help on this one.

Yuck, that sounds messy - I think I have an idea of what's going on, so feel free to make the PR anyway in order to share the error messages. You can always put 'WPI:' at the start of the PR name to show that it is not ready to be merged yet.

PR #320, opened 2020-03-30
port listeners into BaseOperation

### LINKED ISSUE #319: Listener related methods on OptimizeOperation can be ported to BaseOperation
Since #316 we have a `BaseOperation` class  that contains `stop` and `pause` threading events.

However, at the moment, the `EvaluateOperation` class does not set up any `BaseNotificationListener` instances that could influence the `stop` and `pause` events. Therefore we should share the code from `OptimizeOperation` that handles these listener objects, by porting the relevant methods to `BaseOperation` and invoking them from `EvaluateOperation.run`

Overall, we should be left with a `BaseOperation` class containing:

- [ ] `BaseOperation.listeners` attribute
- [ ] `BaseOperation._*_event` methods
- [ ] `BaseOperation._*_listeners` methods

And an `EvaluateOperation.run` method that initialises and finalises the listeners / events before performing its main task

sparsonslab:
Closes #319. Put listener attributes and methods into BaseOperation and tests of these into new test_base_operation.py.


sparsonslab:
For line 43 of `evaluate_operation` ,
`kpi_results = self.workflow.execute(mco_data_values)`
should we put it in a try-except-finally statement like for `mco.run(self.workflow)` in the optimize operation? I tried this, but then the tests were complaining and I couldn't work out what to do.

flongford:
> For line 43 of `evaluate_operation` ,
> `kpi_results = self.workflow.execute(mco_data_values)`
> should we put it in a try-except-finally statement like for `mco.run(self.workflow)` in the optimize operation? I tried this, but then the tests were complaining and I couldn't work out what to do.

That's a good idea, although I think actually this line should be fine if we verify the workflow first (#317), since then the only code that may fail during runtime would just involve the `mco_communicator` instance. In which case it's probably best to put the whole section from line 41-44 in a try-except-finally statement.

The unit tests might complain if log messages are added, so have a go removing the exception logging first. If keeps failing, then it's still ok to push any updates you've made so I can inspect the traceback on the Travis machine.

force_bdss/app/evaluate_operation.py
```diff
@@ -37,5 +41,8 @@ def run(self):
         mco_data_values = mco_communicator.receive_from_mco(mco_model)
 
         kpi_results = self.workflow.execute(mco_data_values)
-
         mco_communicator.send_to_mco(mco_model, kpi_results)
```
flongford: As you suggested, @sparsonslab - I think it would be a good idea to place these lines in a try-except-finally statement, similar to the `OptimizeOperation`

sparsonslab: I have tried issue #317, but see my latest comments.

Whatever one does, it becomes a soup of unit test errors which aren't at all obvious to me.

flongford: I saw in your comment, sounds nasty. 

If this is something you would like to experiment with / keep separate from the current PR then feel free to make a new branch off `base-operation-listeners` with your implementation and then create a PR with it that uses `base-operation-listeners` as the base branch.

That way you can share ideas / ask for review whilst keeping this PR clean.

PR #323, opened 2020-03-31
added workflow verification to evaluate operation.

### LINKED ISSUE #317: EvaluateOperation should verify Workflow first

sparsonslab:
Closes #317. This is not the best solution. 

The problem is, is that `capture.check` (in the unit test) has to match every log in A.foo(), no matter whether it comes from A.foo or BaseOfA.foo(). Therefore you cannot just separate out part of Evaluate/OptimizeOperation.run() into a base class and then test that part separately in base unit test.

What I have done is port the workflow verification part of Evaluate/OptimizeOperation.run() into the base class, but only run that part during testing of OptimizeOperation.run(), by using a do_verify=False switch during testing of EvaluateOperation.run().

### DIFF
--- a/force_bdss/app/evaluate_operation.py
+++ b/force_bdss/app/evaluate_operation.py
@@ -13,13 +13,13 @@ class EvaluateOperation(BaseOperation):
     based on the system described by a `Workflow` object.
     """
 
-    def run(self, do_verify=True):
+    def run(self):
 
         """verify the workflow.
         Do not do use this [run(do_verify=False)] on unit tests :
         use OptimizeOperation.run() to test verification.
         """
-        super().run(do_verify=do_verify)
+        self.verify_workflow()
 
         """ Evaluate the workflow. """
         mco_model = self.workflow.mco_model
--- a/force_bdss/app/evaluate_operation.py
+++ b/force_bdss/app/evaluate_operation.py
@@ -14,19 +14,14 @@ class EvaluateOperation(BaseOperation):
     """
 
     def run(self):
-
-        """verify the workflow.
-        Do not do use this [run(do_verify=False)] on unit tests :
-        use OptimizeOperation.run() to test verification.
+        """ Evaluate the workflow.
         """
+
+        # Verify the workflow
         self.verify_workflow()
 
-        """ Evaluate the workflow. """
+        # optimizer model and factory
         mco_model = self.workflow.mco_model
-        if mco_model is None:
-            log.info("No MCO defined. Nothing to do. Exiting.")
-            return
-
         mco_factory = mco_model.factory
 
         # Set up listeners
--- a/force_bdss/app/evaluate_operation.py
+++ b/force_bdss/app/evaluate_operation.py
@@ -13,7 +13,14 @@ class EvaluateOperation(BaseOperation):
     based on the system described by a `Workflow` object.
     """
 
-    def run(self):
+    def run(self, do_verify=True):
+
+        """verify the workflow.
+        Do not do use this [run(do_verify=False)] on unit tests :
+        use OptimizeOperation.run() to test verification.
+        """
+        super().run(do_verify=do_verify)
+
         """ Evaluate the workflow. """
         mco_model = self.workflow.mco_model
         if mco_model is None:
--- a/force_bdss/app/optimize_operation.py
+++ b/force_bdss/app/optimize_operation.py
@@ -16,7 +16,7 @@ class OptimizeOperation(BaseOperation):
 
     def run(self):
         """ Verify the workflow """
-        super().run(do_verify=True)
+        self.verify_workflow()
 
         """ Create and run the optimizer. """
         mco = self.create_mco()
--- a/force_bdss/app/optimize_operation.py
+++ b/force_bdss/app/optimize_operation.py
@@ -15,10 +15,13 @@ class OptimizeOperation(BaseOperation):
     information during the MCO run."""
 
     def run(self):
-        """ Verify the workflow """
+        """ Create and run the optimizer.
+        """
+
+        # Verify the workflow
         self.verify_workflow()
 
-        """ Create and run the optimizer. """
+        # Create the optimizer
         mco = self.create_mco()
 
         # Set up listeners
--- a/force_bdss/app/optimize_operation.py
+++ b/force_bdss/app/optimize_operation.py
@@ -15,14 +15,10 @@ class OptimizeOperation(BaseOperation):
     information during the MCO run."""
 
     def run(self):
-        """ Create and run the optimizer. """
-        self.workflow_file.verify()
-        if len(self.workflow_file.errors) != 0:
-            log.error("Unable to execute workflow due to verification errors:")
-            for error in self.workflow_file.errors:
-                log.error(error.local_error)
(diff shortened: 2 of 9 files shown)


sparsonslab:
I have made verify workflow a base method and reverted to the old run() signature.

Then there are only two unit test errors in test_evaluate_operation:
 `test_error_for_missing_ds_output_names`
`test_error_for_incorrect_output_slots`

To solve these I figured out in both these test_functions, the workflow file runtime error needed to be added to the with context as an OR.

`        with testfixtures.LogCapture():
            with (self.assertRaisesRegex(
                    RuntimeError,
                    "Workflow file has errors.") or
                  self.assertRaisesRegex(
                    RuntimeError,
                    r"The number of data values returned by the MCO "
                    r"\(1 values\) does not match the number of "
                    r"parameters specified \(0 values\). This is either "
                    "a MCO plugin error or the workflow file is "
                    "corrupted.")):
                self.operation.run() `

sparsonslab:
It passes coverage locally, but I can't get it to pass here. Think its to do with testing log from `if mco_model is None` in the evaluate operation run()

flongford:
> To solve these I figured out in both these test_functions, the workflow file runtime error needed to be added to the with context as an OR.
> 
> ` with testfixtures.LogCapture(): with (self.assertRaisesRegex( RuntimeError, "Workflow file has errors.") or self.assertRaisesRegex( RuntimeError, r"The number of data values returned by the MCO " r"\(1 values\) does not match the number of " r"parameters specified \(0 values\). This is either " "a MCO plugin error or the workflow file is " "corrupted.")): self.operation.run()`

I'm not able to reproduce this - the unit tests appear to work fine with just the first `self.assertRaisesRegex` statement

force_bdss/app/evaluate_operation.py
```diff
@@ -14,6 +14,13 @@ class EvaluateOperation(BaseOperation):
     """
 
     def run(self):
+
+        """verify the workflow.
+        Do not do use this [run(do_verify=False)] on unit tests :
+        use OptimizeOperation.run() to test verification.
+        """
```
flongford: Tidy up / maybe remove this comment?

force_bdss/app/evaluate_operation.py
```diff
@@ -14,6 +14,13 @@ class EvaluateOperation(BaseOperation):
...
+        self.verify_workflow()
+
         """ Evaluate the workflow. """
```
flongford: Style nitpick: we typically use `#` for inline comments, the triple `""" """` quote blocks are reserved for docstrings. This also becomes important for our Sphinx auto doc generation as well

If in doubt, follow the NumPy example code: https://numpydoc.readthedocs.io/en/latest/example.html#source

force_bdss/app/evaluate_operation.py
```diff
@@ -14,6 +14,13 @@ class EvaluateOperation(BaseOperation):
...
         mco_model = self.workflow.mco_model
         if mco_model is None:
```
flongford: Lines 26-28 can be removed, since an error will be raised by `self.verify_workflow` beforehand if no `mco_model` is defined - this might also help the codecov

force_bdss/app/optimize_operation.py
```diff
@@ -15,14 +15,10 @@ class OptimizeOperation(BaseOperation):
...
 
     def run(self):
-        """ Create and run the optimizer. """
-        self.workflow_file.verify()
-        if len(self.workflow_file.errors) != 0:
-            log.error("Unable to execute workflow due to verification errors:")
-            for error in self.workflow_file.errors:
-                log.error(error.local_error)
-            raise RuntimeError("Workflow file has errors.")
+        """ Verify the workflow """
```
flongford: Same style comment as above

force_bdss/app/optimize_operation.py
```diff
@@ -15,14 +15,10 @@ class OptimizeOperation(BaseOperation):
...
+        self.verify_workflow()
 
+        """ Create and run the optimizer. """
```
flongford: Same style comment as above

force_bdss/app/tests/test_evaluate_operation.py
```diff
@@ -100,13 +105,33 @@ def test_error_for_non_matching_mco_parameters(self):
...
+
+        with testfixtures.LogCapture():
+            with (self.assertRaisesRegex(
+                    RuntimeError,
+                    "Workflow file has errors.") or
+                  self.assertRaisesRegex(
+                    RuntimeError,
+                    r"The number of data values \(2 values\)"
+                    " returned by 'test_data_source' does not match"
+                    " the number of user-defined names")):
```
flongford: I wouldn't expect this error statement to be returned - are you able to isolate when it would be raised?

sparsonslab: This is in the master branch and is at least part of a error message raised by ExecutionLayer.execute_layer (line 101, execution_layer.py).
I took it out and coverage works (locally) as well as codecov/patch. But codecov/project fails Whether it is there or not.

sparsonslab: In the latest commit (bc271c0..) I just removed the entire test. As the error message didn't completely match maybe it was vestigial (tests an error no longer there?). There is no equivalent test for the optimize operation.

flongford: I think this happens because in `verify_workflow` we log the individual error messages before raising a new generic error (the `Workflow file has errors` message). Therefore all of the individual errors are no longer going to be reported through `assertRaisesRegex`. In which case we can check the log for these individual messages, rather than the traceback.

The following seems to work for `test_error_for_missing_ds_output_names` (I haven't checked if it improves the coverage)

```python
        with testfixtures.LogCapture() as capture:
            with (self.assertRaisesRegex(
                    RuntimeError,
                    "Workflow file has errors.")):
                self.operation.run()
            capture.check(
                ('force_bdss.app.base_operation',
                 'ERROR',
                 'Unable to execute workflow due to verification errors:'),
                ('force_bdss.app.base_operation',
                 'ERROR',
                 'The number of output slots is incorrect.')
            )
```

flongford: > This is in the master branch and is at least part of a error message raised by ExecutionLayer.execute_layer (line 101, execution_layer.py).

Ohhhh, I see what's going on now. Ok - this is worth a separate PR in it's own right. It looks like there's two places where the input / output slot lengths are being checked - in the `BaseDataSourceModel.verify` method (which is called by `verify_workflow`) and also as you spotted in `ExecutionLayer.execute_layer`.

These means we would get 2 different error messages for the same error, depending on which method raised it - not ideal. We should make sure that only goes on in one place.

PR #325, opened 2020-04-01
added notify_xx_event methods to BaseDataSourceModel

### LINKED ISSUE #318: Use DataSourceStartEvent and DataSourceFinishEvent API in BaseDataSource
Since #315 we have a new API for `DataSourceStartEvent` and `DataSourceFinishEvent` objects.

We can use this in the  [`BaseDataSource._run` ](https://github.com/force-h2020/force-bdss/blob/master/force_bdss/data_sources/base_data_source.py#L29) method, rather than the current implementation. Doing so will require creating `notify_start_event` and `notify_finish_event` on the `BaseDataSourceModel` class (again, following the `BaseMCOModel` example).

# Note:
Creating unit tests for this part might be a little bit fiddly, but looking at https://github.com/force-h2020/force-bdss/blob/master/force_bdss/mco/tests/test_base_mco_model.py#L67 should probably help

sparsonslab:
Closes #318 (hopefully!). I think the test I added does something useful.


sparsonslab:
> This is all looking good so far, many thanks
> 
> The last step to close #318 will be to edit the `BaseDataSource._run` method to use this new code - something like:
> 
> ```python
> model.notify_start_event()
> result = self.run(model, parameters)
> model.notify_finish_event()
> ```

It already does this in the private run(). It should be implemented in abstract run() as well?

PR #326, opened 2020-04-01
change local error message for input/output verification

### LINKED ISSUE #324: Different error messages reported for same Input/Output slot length verification
As identified in #323, the length of input and output slot lists are being verified in two different places:

### Original discussion:

> This is in the master branch and is at least part of a error message raised by ExecutionLayer.execute_layer (line 101, execution_layer.py).

Ohhhh, I see what's going on now. Ok - this is worth a separate PR in its own right. It looks like there's two places where the input / output slot lengths are being checked - in the `BaseDataSourceModel.verify` method (which is called by `verify_workflow`) and also as you spotted in `ExecutionLayer.execute_layer` (which would have been reported back if we hadn't called `verify_workflow` first).

These means we would get 2 different error messages for the same error, depending on which method raised it - not ideal. We should make sure that this only happens in one place. I will make a new issue.

_Originally posted by @flongford in https://github.com/force-h2020/force-bdss/pull/323_

sparsonslab:
Closes #324. Changed local error message for check on no. of inputs/outputs, in `BaseModelDataSource.verify()`. Message conforms with verification in `ExecutionLayer.execute_layer()` (corresponding parts of this now removed).

Corresponding changes to unit tests (includes fixes from #327).


sparsonslab:
codecov/project is weird, right??

flongford:
> codecov/project is weird, right??
> 
> It just says "everything is accounted for"

Yes... it can be very unhelpful at times. The only line I can see in `BaseDataSourceModel` that is not being covered is line 95 (`raise` exception if `BaseDataSource.slots` method fails), which is not touched by this PR.

If you were able to make a quick test for this, then it will likely stop complaining.

flongford:
> Having said that, it looks like there IS already a test in `test_base_data_source_model.py` doing just that!

Ok, after so experimenting a bit I think the `TestBaseDataSourceModel.test_bad_slots` method is not working properly. There might be a bit to much `mock`ing being used - I will see if I can simplify things

PR #327, opened 2020-04-01
TST/FIX: Suggested fixes to TestBaseDataSourceModel class

flongford:
Includes fixes to mocked objects that ensure both `TestBaseDataSourceModel.test_bad_factory` and `TestBaseDataSourceModel.test_bad_slots` are asserting that the correct `Exceptions` have been raised.

Previously, an error was being generated in the invocation of `self.mock_factory.create_data_source`, related to a missing `plugin_id` attribute on the mocked `BaseDataSourceFactory` class. This then prevented any further code from being tested.

This has now been resolved, with `testfixtures.LogCapture` used to check the logs and tidy up the unit test traceback.

## Note:
I couldn't get the `capture.check` call to work in `test_bad_slots` (string formatting I expect), so I left it out. The `LogCapture` context manager is still used, since it prevents the log statements from cluttering up the unit test output.


sparsonslab:
Ha, I remember that extra space in "a programming"! I was getting lost in the same code.

PR #328, opened 2020-04-01
Verification message two

### LINKED ISSUE #324: Different error messages reported for same Input/Output slot length verification

sparsonslab:
Closes #324. 

Ported output/input slot error messages from ExecutionLayer.execute_layer() to BaseDataSourceModel.verify().

Did not remove tests from execute_layer() - that took me down a unit-test rabbit hole.

### DIFF
--- a/force_bdss/app/tests/test_optimize_operation.py
+++ b/force_bdss/app/tests/test_optimize_operation.py
@@ -139,10 +139,6 @@ class TestOptimizeOperation(TestCase):
                 self.operation._deliver_start_event()
             mock_call.assert_called_once()
 
-    ##############################################
-    # RUN TESTS: POSSIBLY COMMON WITH EVALUATE OPERATION
-    ##############################################
-
     def test_non_valid_file(self):
 
         # Provide a workflow that is invalid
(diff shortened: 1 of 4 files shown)


flongford:
@sparsonslab - Did you see #327?

sparsonslab:
> @sparsonslab - Did you see #327?

No I hadn't. I can incorporate that into this branch. (?)

flongford:
> > @sparsonslab - Did you see #327?
> 
> No I hadn't. I can incorporate that into this branch. (?)

I think it should already solve the issues you had with #326

flongford:
Closing in preference of #326

Issue #321, opened 2020-03-30
WeightedOptimizerEngine could inherit from ScipyOptimizerEngine


flongford:
At the moment, we offer the `WeightedOptimizerEngine` class as a useful object to be used in any `BaseMCO.run` implementation. However, it is currently providing two key functionalities, which could be separated: an entry point to the `scipy.optimizers` package, and the ability to assign importance weights to KPIs.

It may be beneficial to create a new subclass of `BaseOptimizerEngine` that only exposes the `scipy.optimizer` routines, which `WeightedOptimizerEngine` would then inherit from. ~By doing so, we may also explore the SciPy 'ask and tell' interface, similar to our Nevergrad plugin classes.~

Note: as of writing (8/04/20) the SciPy library does not have an "ask and tell" API for its optimiser module.

Solving this issue will require the following:
- [x] A new class: `SciPyOptimizerEngine`, inheriting from `BaseOptimizerEngine`. This should fulfil the base class API only and act as an entry point for the `scipy.optimizer` library. Much of the current required functionality is probably contained in the `WeightedOptimizerEngine._weighted_optimize` method.

- [ ] Refactoring `WeightedOptimizerEngine` to inherit from `SciPyOptimizerEngine`. This could be done in a number of ways, but we should focus on keeping the weighted features independent from the optimization library being used. A `WeightedKPISearch` mixin class might therefore be appropriate, which could also be used with any other `BaseOptimizerEngine` subclass

flongford:
We could handle KPI weights in a different way - maybe as attributes of the `KPISpecification` class?

sparsonslab:
Things the user might want to do, that could be implemented as a generic MCO "Model" class (i.e. Traits view). 

* Sample (the objective function)
	* random
	* grid
* Optimize (find the minima/maxima of the objective function)
	* what optimizer I want to use. Each one should be associated with a implementation of the 	BaseOptimizerEngine class.
		* scipy (just gradient)
		* scikit (Bayesian, forest regression, etc)
		* nevergrad (many!)
	* what I want to do with the optimizer. These could be collected together as an interface that each OptimizerEngine has to 	implement. e.g. optimize_single_x0().
		* from single x0 (search for single minimum)
		* from random x0 (search for all minima)
		* Pareto front/set
			* A priori methods (can be applied using any optimizer)
				* scalarisation (i.e. like "weighted optimizer")
			* A posteriori methods (only implemented in some packages: e.g. nevergrad).
[...]

PR #340, opened 2020-04-23
separated scipy from weights

sparsonslab:
## Description
Contributes to #321. Separates scipy optimization from the consideration of weights, by creating a new `ScipyOptimizerEngine` class whose `optimize()` method just returns the optimal point found. `WeightedOptimizerEngine` then becomes a subclass of `ScipyOptimizerEngine`.

### Summary
Separates scipy optimization from the consideration of weights.

### Future Work
To create a mixin-class for weighted methods (as suggested in #321)

## Changes
Added `ScipyOptimizerEngine` class. `WeightedOptimizerEngine` is much the same as it was before,
but now:
- inherits from `ScipyOptimizerEngine` rather than `BaseOptimizerEngine` directly.
- `WeightedOptimizerEngine._weighted_optimize()` calls `ScipyOptimizerEngine._scipy_optimize()` rather than implementing the call to scipy's minimize itself.

- _This section will also be used as a reference to construct the `CHANGES.rst` entry for each release_


force_bdss/mco/optimizer_engines/space_sampling.py
```diff
@@ -40,12 +40,6 @@ class SpaceSampler(ABCHasStrictTraits):
     def __init__(self, dimension, resolution, **kwargs):
         super().__init__(dimension=dimension, resolution=resolution, **kwargs)
 
-    @abc.abstractmethod
```
sparsonslab, line 43: I didn't intend to do this. I did pull from master to begin, so I don't know whats happening with this.

force_bdss/mco/optimizer_engines/scipy_optimizer_engine.py
```diff
@@ -0,0 +1,94 @@
...
+            Point of evaluation, objective value, dummy list of weights
+        """
+        #: Get scaling factors and non-zero weight combinations for each KPI
+        scaling_factors = self.get_scaling_factors()
+
+        optimal_point, optimal_kpis = self._scipy_optimize(
+            self._score
+        )
+        yield optimal_point, optimal_kpis, scaling_factors
+
```
sparsonslab, line 62: the weighted optimizer will call _scipy_optimize() with the partial of _score containing the weights

force_bdss/mco/optimizer_engines/scipy_optimizer_engine.py
```diff
@@ -0,0 +1,94 @@
...
+    parameter_bounds = Property(
+        depends_on="parameters.[lower_bound, upper_bound]", visible=False
+    )
+
+    def _get_initial_parameter_value(self):
+        return [p.initial_value for p in self.parameters]
+
+    def _get_parameter_bounds(self):
+        return [(p.lower_bound, p.upper_bound) for p in self.parameters]
+
```
sparsonslab, line 46: just return the single (unweighted) minima. overridden in the weighted subclass

force_bdss/mco/optimizer_engines/weighted_optimizer_engine.py
```diff
@@ -146,21 +125,7 @@ def _weighted_optimize(self, weights):
 
         weighted_score_func = partial(self.weighted_score, weights=weights)
```
sparsonslab, line 127: _scipy_optimize now carrys out the following removed code

force_bdss/mco/optimizer_engines/scipy_optimizer_engine.py
```diff
@@ -0,0 +1,98 @@
...
+            self._scalar_score
+        )
+        yield optimal_point, optimal_kpis
+
+    def _scalar_score(self, input_point):
+        """ The sum of the objectives/kpis, so they can be
+        optimized by scipy.
+        """
+        return np.sum(self._score(input_point))
+
```
sparsonslab, line 66: The scipy optimizer can only optimizer a scalar function, so we still have to add up the objectives.

force_bdss/mco/optimizer_engines/scipy_optimizer_engine.py
```diff
@@ -0,0 +1,98 @@
...
+        Yields
+        ----------
+        optimization result: tuple(np.array, np.array, list)
+            Point of evaluation, objective value, dummy list of weights
+        """
+        optimal_point, optimal_kpis = self._scipy_optimize(
+            self._scalar_score
+        )
+        yield optimal_point, optimal_kpis
+
```
sparsonslab, line 60: Just returns the single minimum found.

force_bdss/mco/optimizer_engines/scipy_optimizer_engine.py
```diff
@@ -0,0 +1,98 @@
...
+        )
+        optimal_point = optimization_result.x
+        optimal_kpis = self._score(optimal_point)
+
+        log.info(
+            "Optimal point : {}".format(optimal_point)
+            + "KPIs at optimal point : {}".format(optimal_kpis)
+        )
+
+        return optimal_point, optimal_kpis
```
sparsonslab, line 98: This called by the weighted-optimizer-engine subclass with the partial of _score with the weights applied.

force_bdss/mco/optimizer_engines/space_sampling.py
```diff
@@ -40,12 +40,6 @@ class SpaceSampler(ABCHasStrictTraits):
...
-    def _get_sample_point(self):
-        """ Internal generator of sample points from the distribution.
-        Returns or yields a sampled point.
-        """
-
```
sparsonslab, line 48: I didn't make any changes to this module and I 'm pretty sure I pulled from master, so I'm not sure what's going on here. But it doesn't make any difference to running.

force_bdss/mco/optimizer_engines/weighted_optimizer_engine.py
```diff
@@ -66,34 +67,12 @@ class WeightedOptimizerEngine(BaseOptimizerEngine):
     #: Search grid resolution per KPI
     num_points = PositiveInt(7)
 
-    #: Algorithms available to work with
-    algorithms = Enum("SLSQP", "Nelder-Mead", "Powell", "CG", "BFGS",
-                      "Newton-CG", "L-BFGS-B", "TNC", "COBYLA",
-                      "trust-constr", "dogleg",
-                      "trust-ncg", "trust-exact", "trust-krylov")
-
```
sparsonslab, line 74: move out all the attributes related to scipy.....

force_bdss/mco/optimizer_engines/weighted_optimizer_engine.py
```diff
@@ -146,21 +125,7 @@ def _weighted_optimize(self, weights):
...
-        )
-        optimal_point = optimization_result.x
-        optimal_kpis = self._score(optimal_point)
-
-        log.info(
-            "Optimal point : {}".format(optimal_point)
-            + "KPIs at optimal point : {}".format(optimal_kpis)
-        )
-
-        return optimal_point, optimal_kpis
```
sparsonslab, line 163: Now just call _scipy_optimizer which contains all this code.

sparsonslab:
Obviously doesn't pass code coverage, as I haven't put in any tests for the new scipy-engine class. But is this the kind of thing you were thinking of?

force_bdss/mco/optimizer_engines/space_sampling.py
flongford: Yes, the diff will change if any code is moved around, the functionality looks the same

force_bdss/mco/optimizer_engines/scipy_optimizer_engine.py
flongford: Would it be possible to refactor `WeightedOptimizerEngine.optimize` to call this as a `super` method (rather than overwriting completely)? If that's possible, then we could go on to make the weighted-scale-search components sit on top of any `BaseOptimizerEngine` implementation.

Note: This will probably require significant refactoring of `WeightedOptimizerEngine`, so is may be worth doing in a separate PR.

sparsonslab: You would need to able to tell the super that you want to use the weighted partial, but as optimize() has no arguments, other than self, there is no way to do this.

flongford: Agreed - I can't think of a way to do this without changing the `BaseOptimizerEngine`. Let's leave this for now then.

Issue #342, opened 2020-04-24
Explore making WeightedOptimizerEngine class a mixin


flongford:
We can explore ways to make the `WeightedOptimizerEngine` class not rely on inheriting from `ScipyOptmizerEgine`, and rather be a mixin-type object that can sit on top of any `BaseOptimizerEngine` implementation.

The mixin will need to:

- [ ] Overwrite the objective function to include weights on each KPI value returned
- [ ] Overload the `optimize` method to iterate through each combination of weights returned by the space sampler, calling the `super().optimize` method in turn.

flongford:
Since #340, the only dependency that `WeightedOptimizerEngine` has on `SciPyOptimizerEngine` is via use of the `_scipy_optimize` method in `_weighted_optimize`.

We would need to invoke the `super().optimize` method here (or equivalent) for this to be independent of the parent class. However, the if there are any extra functionalities in `_scipy_optimize` that are missing from `BaseOptimizerEngine`, then we will also need to ensure these are present in the base class.

sparsonslab:
> We can explore ways to make the `WeightedOptimizerEngine` class not rely on inheriting from `ScipyOptmizerEgine`, and rather be a mixin-type object that can sit on top of any `BaseOptimizerEngine` implementation.
> 
> The mixin will need to:
> 
> * [ ]  Overwrite the objective function to include weights on each KPI value returned
> * [ ]  Overload the `optimize` method to iterate through each combination of weights returned by the space sampler, calling the `super().optimize` method in turn.

Would the goal be something like,

`class MyOptimizerEngine(ScipyOptimizerEngine, WeightedOptimizerEngine):
pass`

where you don't have to implement any methods in MyOptimizerEngine - it just inherits from its bases (ScipyOptimizerEngine and WeightedOptimizerEngine). In that case those bases should be unrelated: only one inherits from BaseOptimizerEngine and provides the optimize(). The other provides some other interface that optimize calls on.

Does that make sense?

flongford:
> Would the goal be something like,
> 
> `class MyOptimizerEngine(ScipyOptimizerEngine, WeightedOptimizerEngine): pass`
> 
> where you don't have to implement any methods in MyOptimizerEngine - it just inherits from its bases (ScipyOptimizerEngine and WeightedOptimizerEngine).

Yes, that would be the idea - or a namespace equivalent:

```python
class WeightedOptimizerEngine(ScipyOptimizerEngine, WeightedOptimizerMixin):
[...]

PR #343, opened 2020-04-24
optimizer is a mixin with an interface

sparsonslab:
### Description
We want an easy way to create child of `BaseOptimizerEngine` ("`MyOptimizerEngine`")
that can mix-and-match:
- optimizer (scipy, nevergrad, sci-kit, etc.) and 
- optimization context (weighted-multiobjective, pareto-front, etc.)

To do this `MyOptimizerEngine` handles the context (as before), but also inherits an mixin class
that contributes the optimizer in the form of:
- an optimize_function() method. This is imposed on the mixin, by an interface
- attributes associated with that optimizer, such as algorithm names.

When you want to change the optimizer, you simply create a child of `MyOptimizerEngine` that
also inherits a different optimizer class: as long as this optimizer is first in the inheritance signature, it will override that of `MyOptimizerEngine`.

To illustrate the following toy code:
```
class optimizer_a:
    def get_name(self):
        return 'a'
    
class optimizer_b:
    def get_name(self):
        return 'b'
    
class engine_j(optimizer_a):
    
    def get_optimizser_name(self):
        return 'optimizer = ' + self.get_name()

class engine_q(optimizer_b, engine_j):
    pass

print(engine_j().get_optimizser_name())

print(engine_q().get_optimizser_name())
```

```
>> optimizer = a
>> optimizer = b
```

### Related Issues
#342 

## Changes
- ScipyOptimizer is now not an engine, but rather just implements an optimizer interface (IOptimizer) that requires the _optimize_function() method to be called by the engine.
- WeightedOptimizerEngine now inherits both BaseOptimizerEngine and ScipyOptimizer.

### Future Work
Create optimizers other than ScipyOptimize, that also implement the IOptimizer interface.
Create a subclass of WeightedOptimizerEngine that inherits a different optimizer.


force_bdss/mco/optimizer_engines/i_optimizer.py
```diff
@@ -0,0 +1,25 @@
...
+    @provides(IOptimizer)
+    class ScipyOptimizer():
+
+    Any subclass of BaseOptimizerEngine can inherit such a class
+    and then call optimize_function() within its optimize().
+    """
+    @abc.abstractmethod
+    def optimize_function(self, func, x0, bounds):
+        """ Optimize a function, according to the library (scipy/etc).
+        """
```
sparsonslab, line 25: optimize_function() is called from within the engine. I had some more on the docstring but the doc build was complaining.

force_bdss/mco/optimizer_engines/weighted_optimizer_engine.py
```diff
@@ -55,15 +56,23 @@ def sen_scaling_method(dimension, weighted_optimize):
     return scaling_factors
 
 
-class WeightedOptimizerEngine(ScipyOptimizerEngine):
-    """ Performs local optimization of multiobjective function using scipy.
-    The multiobjective function is converted to a scalar by dot product
-    with a weights vector (`weighted_score`).
+class WeightedOptimizerEngine(BaseOptimizerEngine, ScipyOptimizer):
+    """ Performs multi-objective optimization by weighting.
     """
```
sparsonslab, line 62: I had some extensive docstring here, but the doc build complained so I took it out ;(

force_bdss/mco/optimizer_engines/weighted_optimizer_engine.py
```diff
@@ -125,7 +142,19 @@ def _weighted_optimize(self, weights):
...
         weighted_score_func = partial(self.weighted_score, weights=weights)
 
-        return self._scipy_optimize(weighted_score_func)
+        # use default .optimize_function
+        optimal_point = self.optimize_function(
+            weighted_score_func,
+            self.initial_parameter_value,
+            self.parameter_bounds)
+        optimal_kpis = self._score(optimal_point)
+
```
sparsonslab, line 151: Here we call the ScipyOptimize.optimize_function(). If we create a child of WeightedOptimizerEngine that also inherits say NevergradOptimize as the *first* inheritance arg, it will call NevergradOptimize.optimize_function() instead.

force_bdss/mco/optimizer_engines/weighted_optimizer_engine.py
flongford: This might be due to the use of underscores to signify a title word that isn't recognised (Sphinx expects specific title words like `Parameters` and `Returns` in docstrings). It looks like the doc build is still complaining for `i_optimizer.py` because of this

Try using ~~~~~ to underscore the test instead, rather than -------

force_bdss/mco/optimizer_engines/weighted_optimizer_engine.py
```diff
@@ -55,15 +56,23 @@ def sen_scaling_method(dimension, weighted_optimize):
     return scaling_factors
 
 
-class WeightedOptimizerEngine(ScipyOptimizerEngine):
-    """ Performs local optimization of multiobjective function using scipy.
-    The multiobjective function is converted to a scalar by dot product
-    with a weights vector (`weighted_score`).
+class WeightedOptimizerEngine(BaseOptimizerEngine, ScipyOptimizer):
```
flongford, line 59: Can we define `WeightedOptimizerEngine` independently of `ScipyOptimizer`? Otherwise we won't be able to swap the `IOptimizer` mixin for something else.

flongford: (This links to the comment above)

force_bdss/mco/optimizer_engines/weighted_optimizer_engine.py
sparsonslab: Alternative
```
class optimizer_a:
    def get_name(self):
        return 'a'
    
class optimizer_b:
    def get_name(self):
        return 'b'
    
class engine_j():
    
    def get_name(self):
        """ Dummy function (or leave out in the expectation...)"""
        return 'none'
    
    def get_optimizser_name(self):
        return 'optimizer = ' + self.get_name()

class engine_q(optimizer_b, engine_j):
    pass

print(engine_j().get_optimizser_name())

print(engine_q().get_optimizser_name())
```

sparsonslab: > Have you experimented with implementing an appropriate `NevergradOptimize.optimize_function` method? Is this an appropriate signature for both external libraries?

Yes

sparsonslab: We can dynamically create engine/optimizer mixes (in reference to a user selection).

```
myengine = type('MyEngine', (optimizer_b, engine_j), {})
print(myengine().get_optimizser_name())
```

force_bdss/mco/optimizer_engines/weighted_optimizer_engine.py
sparsonslab: Of course - you can just leave ScipyOptimizer out of the inheritence. As long as you don't try to then make an instance of the class and run it (it will call a method that isn't defined).

force_bdss/mco/optimizer_engines/weighted_optimizer_engine.py
sparsonslab: ... the only thing with the dummy method, is that you still end up overriding it when you inherit an optimizer, so it really doesn't make any difference (to overiding an inheritance from another optimizer).

force_bdss/mco/optimizer_engines/weighted_optimizer_engine.py
sparsonslab: Or you could replace ScipyOptimizer by some BaseOptimizer, that just has an optimize_function method that returns [1, 1, 1] or something.

sparsonslab: I don't think this is matter or the right-way or the wrong-way in terms of python. There are dozens of possibilities that are all perfectly valid python. The choice is a matter the wider spec (how these classes are going to be created and used) and taste.

sparsonslab: > Or you could replace ScipyOptimizer by some BaseOptimizer, that just has an optimize_function method that returns [1, 1, 1] or something.

This would be good, because the engine is only inheriting the minimal methods (optimize_function) and attributes (none!) which are guaranteed to be overwritten by inherited another optimizer.

force_bdss/mco/optimizer_engines/i_optimizer.py
```diff
@@ -0,0 +1,25 @@
...
+    An implementing class will be specific to an optimizer library. e.g.
+
+    @provides(IOptimizer)
+    class ScipyOptimizer():
+
+    Any subclass of BaseOptimizerEngine can inherit such a class
+    and then call optimize_function() within its optimize().
+    """
+    @abc.abstractmethod
+    def optimize_function(self, func, x0, bounds):
```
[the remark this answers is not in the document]
flongford: Actually, these arguments are all required by the `WeightedOptimizerEngine`, it makes sense to include `x0, bounds`.

force_bdss/mco/optimizer_engines/weighted_optimizer_engine.py
flongford: Could we put `optimize_function` as an abstract method on the `WeightedOptimizerEngine`?

force_bdss/mco/optimizer_engines/i_optimizer.py
```diff
@@ -0,0 +1,15 @@
+import abc
```
flongford, line 1: Let's import this  class in the `force_bdss.api` module so it can be used by other plugins

force_bdss/mco/optimizer_engines/weighted_optimizer_engine.py
```diff
@@ -55,15 +56,23 @@ def sen_scaling_method(dimension, weighted_optimize):
...
+    #: Default (initial) guess on input parameter values
+    initial_parameter_value = Property(
+        depends_on="parameters.[initial_value]", visible=False
+    )
+
+    #: Input parameter bounds. Defines the search space.
+    parameter_bounds = Property(
+        depends_on="parameters.[lower_bound, upper_bound]", visible=False
+    )
+
```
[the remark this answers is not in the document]
flongford: Actually these seem more relevant for the `WeightedOptimizerEngine`, so we can leave them where they are for the time being.

sparsonslab:
I'll need some help with this. It appears the build can't get hold of nevergrad. Does nevergrad need to be installed somewhere?

sparsonslab:
..so nevergrad has to be in `PIP_DEPS` not `ADDITIONAL_PIP_DEPS`.

flongford:
> ..so nevergrad has to be in `PIP_DEPS` not `ADDITIONAL_PIP_DEPS`.

We should not import Nevergrad here, it is not part of the core framework. If you are testing objects with dependencies on Nevergrad, then contribute them as a PR in the `force-bdss-plugin-nevergrad`

force_bdss/mco/optimizers/nevergrad_optimizers.py
```diff
@@ -0,0 +1,139 @@
+from traits.api import (
```
flongford, line 1: This code should be in the `force-bdss-plugin-nevergrad` repo. Could you please make a new PR for it there?

sparsonslab: The PR is there (#17) but won't build because it has bdss dependencies that aren't in the master.

force_bdss/mco/optimizers/scipy_optimizer.py
```diff
@@ -0,0 +1,35 @@
...
+    """ Optimization of a scalar function using scipy.
+    """
+
+    #: Algorithms available to work with
+    algorithms = Enum("SLSQP", "Nelder-Mead", "Powell", "CG", "BFGS",
+                      "Newton-CG", "L-BFGS-B", "TNC", "COBYLA",
+                      "trust-constr", "dogleg",
+                      "trust-ncg", "trust-exact", "trust-krylov")
+
+    def optimize_function(self, func: Callable, x0: list, bounds: list):
```
flongford, line 24: Can we put the parameter types in a docstring here, rather than in the signature?

sparsonslab: yeah

force_bdss/mco/optimizer_engines/aposteriori_optimizer_engine.py
```diff
@@ -0,0 +1,49 @@
...
+        """
+        #: get pareto set
+        pareto = self.optimizer.optimize_function(
+            self._score,
+            self.initial_parameter_value,
+            self.parameter_bounds
+        )
+        for optimal_point in pareto:
+            optimal_kpis = self._score(optimal_point)
+            yield optimal_point, optimal_kpis
```
flongford, line 45: **Note:** In this design, the `optimize_function` has to calculate and store all points initially, before they are yielded. Therefore our UI will not be updated with each point as it comes in. If we allow for `optimize_function` to be implemented as a generator that will yield `optimal_point`, then we can achieve this by putting in place something like:

```python
    for optimal_point in self.optimizer.optimize_function(
            self._score,
            self.initial_parameter_value,
            self.parameter_bounds
        )
         optimal_kpis = self._score(optimal_point)
         yield optimal_point, optimal_kpis
```

force_bdss/mco/optimizer_engines/weighted_optimizer_engine.py
```diff
@@ -123,9 +146,29 @@ def _weighted_optimize(self, weights):
...
+            self.initial_parameter_value,
+            self.parameter_bounds)
+        optimal_kpis = self._score(optimal_point)
+
+        log.info(
+            "Optimal point : {}".format(optimal_point)
+            + "KPIs at optimal point : {}".format(optimal_kpis)
+        )
+
+        return optimal_point, optimal_kpis
```
flongford, line 164: This code could be rewritten if `optimize_function` is implemented as a generator yielding `optimal_point`:

```python
    for optimal_point in self.optimizer.optimize_function(
            weighted_score_func,
            self.initial_parameter_value,
            self.parameter_bounds)

        optimal_kpis = self._score(optimal_point)

        log.info(
            "Optimal point : {}".format(optimal_point)
            + "KPIs at optimal point : {}".format(optimal_kpis)
        )

        yield optimal_point, optimal_kpis
```

force_bdss/mco/optimizer_engines/weighted_optimizer_engine.py
```diff
@@ -101,8 +116,16 @@ def optimize(self):
             optimal_point, optimal_kpis = self._weighted_optimize(
                 scaled_weights
             )
+
             yield optimal_point, optimal_kpis, scaled_weights
```
flongford, line 120: This then becomes:

```python
for optimal_point, optimal_kpis in self._weighted_optimize(
                scaled_weights
            ):
    yield optimal_point, optimal_kpis, scaled_weights
```

sparsonslab:
I have also modified the itwm example so that it creates a weighted-engine with the scipy optimizer instance (rather than mixin). Shall I push that as well? It works in both bdss and wfmanager.

sparsonslab:
Is there any reason why 
`class RangedVectorMCOParameter(RangedMCOParameter):`

is not derived from BaseMCOParameter instead? It's confusing my unit tests because the scipy optimizer is trying to distinguish `RangedMCOParameter` from `RangedVectorMCOParameter` based on isinstance() test

sparsonslab:
> Is there any reason why
> `class RangedVectorMCOParameter(RangedMCOParameter):`
> 
> is not derived from BaseMCOParameter instead? It's confusing my unit tests because the scipy optimizer is trying to distinguish `RangedMCOParameter` from `RangedVectorMCOParameter` based on isinstance() test

I just changed the code so that the instance test for `RangedVectorMCOParameter` comes before that for `RangedMCOParameter` in the if-else tree.

flongford:
> I have also modified the itwm example so that it creates a weighted-engine with the scipy optimizer instance (rather than mixin). Shall I push that as well? It works in both bdss and wfmanager.

Yes please go ahead and create a PR in the `force-bdss-plugin-itwm-example`

force_bdss/mco/optimizer_engines/tests/test_aposteriori_optimizer_engine.py
```diff
@@ -0,0 +1,77 @@
...
+            },
+            state,
+        )
+
+    def test_optimize(self):
+        """ To test this we need an a posteriori optimizer installed in bdss.
+        """
+        # NOTE: we can use a dummy IOptimize class here to check the
+        # values that should be returned from calling
+        # AposterioriOptimizerEngine.optimize
```
flongford, line 77: Making a `ProbeOptimizer` (`IOptimizer`) class that returns a predictable set of values for `optimizer_func` method will help here (we may also need a `ProbeEvaluator` class as well to get `_score` working.

force_bdss/tests/probe_classes/i_evaluator.py
```diff
@@ -0,0 +1,21 @@
...
+
+from force_bdss.mco.base_mco_model import BaseMCOModel
+
+
+class DummyMCOModel(BaseMCOModel):
+    pass
+
+
+@provides(IEvaluator)
+class ProbeIEvaluator:
```
flongford, line 16: Nitpick: this is not an `Interface` class, so the naming convention would be to drop the `I` here and call it `ProbeEvaluator`.

Similarly, this file should probably be called `evaluator.py` rather than `i_evaluator.py`

force_bdss/tests/probe_classes/i_optimizer.py
```diff
@@ -0,0 +1,36 @@
...
+from force_bdss.mco.parameters.mco_parameters import (
+    FixedMCOParameter,
+    RangedMCOParameter,
+    ListedMCOParameter,
+    CategoricalMCOParameter
+)
+
+
+@provides(IOptimizer)
+class ProbeIOptimizer(HasStrictTraits):
```
flongford, line 19: Nitpick: (same as above)

force_bdss/api.py
```diff
@@ -37,13 +37,13 @@
...
 from .mco.i_evaluator import IEvaluator  # noqa
 from .mco.i_mco_factory import IMCOFactory  # noqa
+from .mco.optimizers.i_optimizer import IOptimizer  # noqa
 from .mco.parameters.base_mco_parameter_factory import BaseMCOParameterFactory  # noqa
 from .mco.parameters.base_mco_parameter import BaseMCOParameter  # noqa
 from .mco.parameters.mco_parameters import FixedMCOParameterFactory, RangedMCOParameterFactory, ListedMCOParameterFactory, CategoricalMCOParameterFactory, RangedVectorMCOParameterFactory  # noqa
 from .mco.parameters.mco_parameters import FixedMCOParameter, RangedMCOParameter, ListedMCOParameter, CategoricalMCOParameter, RangedVectorMCOParameter  # noqa
 from .mco.optimizer_engines.base_optimizer_engine import BaseOptimizerEngine  # noqa
 from .mco.optimizer_engines.weighted_optimizer_engine import WeightedOptimizerEngine  # noqa
-from .mco.optimizer_engines.scipy_optimizer_engine import ScipyOptimizerEngine  # noqa
```
flongford, line 46: We could expose `ScipyOptimizer` here instead as it will be required by the ITWM plugin

force_bdss/mco/optimizers/scipy_optimizer.py
```diff
@@ -0,0 +1,265 @@
...
+                      "trust-ncg", "trust-exact", "trust-krylov")
+
+    def optimize_function(self, func, params):
+        """ Minimize the passed function.
+
+        Parameters
+        ----------
+        func: Callable
+            The MCO function to optimize
+            Takes a list of MCO parameter values.
```
flongford, line 41: Maybe include 'must return a scalar value', if we only support functions that can be optimised by `scipy.optimize.minimize`. Alternatively we could state that any non-scalar value will be converted using a summation (see next comment).

force_bdss/mco/optimizers/scipy_optimizer.py
```diff
@@ -0,0 +1,265 @@
...
+        """
+
+        # Translate the numpy array into an MCO parameter list
+        param_values = self.translate_array_to_mco(array, params)
+
+        # Call the function that takes a list of MCO parameter values
+        objective = func(param_values)
+
+        # Translate the objective (kpis) into a scalar if it is not.
+        # ??????????
```
flongford, line 107: Was this unfinished? Would something like:

```python
if not np.isscalar(objective):
    objective = np.sum(objective)
```

be suitable (if this behaviour is documented)?

PR #344, opened 2020-04-28
DEV: Suggested Edits for #343

flongford:
### Related Issues
#342 #343 

### Background
Suggested usage of `IOptimizer` classes - although #342 originally planned to have a mixin based system, it may make better sense to use `Interface` classes instead.

### Summary
Rather than using these classes as mixins for `BaseOptimizerEngine` subclasses, we include them as traits. 

This also allows for reallocation of a library backend during runtime.

### Notes:
I am sure there are some confusing naming conventions going on here, happy to hear any alternatives

### Future Work
- Update existing PRs for Nevergrad and ITWM MCO engines
- Expose `IOptimizer` classes in the UI

## Changes
- Moves `initial_parameter_value` and `parameter_bounds` to `BaseOptimizerEngine`
- Includes `optimizer` attribute on `WeightedOptimizerEngine` and `APosterioriEngine` classes, which must fulfil `IOptimizer` interface
- Removes `optimize_function` method from all `BaseOptimizerEngine` subclasses, defers usage to `self.optimizer.optimize_function` instead

### DIFF
--- a/force_bdss/mco/optimizer_engines/aposteriori_optimizer_engine.py
+++ b/force_bdss/mco/optimizer_engines/aposteriori_optimizer_engine.py
@@ -24,7 +24,7 @@ class AposterioriOptimizerEngine(BaseOptimizerEngine):
 
     #: IOptimizer class that provides library backend for optimizing a
     #: callable
-    optimizer = Instance(IOptimizer)
+    optimizer = Instance(IOptimizer, transient=True)
 
     def optimize(self, *vargs):
         """ Generates optimization results.
--- a/force_bdss/mco/optimizer_engines/aposteriori_optimizer_engine.py
+++ b/force_bdss/mco/optimizer_engines/aposteriori_optimizer_engine.py
@@ -1,7 +1,8 @@
 import logging
-import abc
 
-from traits.api import Str, Property
+from traits.api import Str, Instance
+
+from force_bdss.mco.optimizers.i_optimizer import IOptimizer
 
 from .base_optimizer_engine import BaseOptimizerEngine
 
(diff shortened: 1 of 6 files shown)


flongford:
Note: this is failing on coverage only, I am going to merge and then we can work with this in #343

sparsonslab:
Bear in mind, that by making the optimizer an instance of the engine, rather than a mixin, it will change the way the optimizer's attributes (algorithms, etc) are exposed to any `MCO` or `MCOModel`. With the mixin, those attributes are inherited by the engine and so just appear like they used to, e.g. engine.algorithms. Now you have engine.optimizer.algorithms.

flongford:
> Bear in mind, that by making the optimizer an instance of the engine, rather than a mixin, it will change the way the optimizer's attributes (algorithms, etc) are exposed to any `MCO` or `MCOModel`. With the mixin, those attributes are inherited by the engine and so just appear like they used to, e.g. engine.algorithms. Now you have engine.optimizer.algorithms.

This is true, we will need to address this when deciding on how to fit the `BaseOptimizerEngine` and `IOptimizer` classes into the BDSS UI framework.

sparsonslab:
I've also realised we don't need to pass x0 and bounds as args to optimize_function. All we need to pass is self.parameters, as these contain both x0 and bounds.

flongford:
> I've also realised we don't need to pass x0 and bounds as args to optimize_function. All we need to pass is self.parameters, as these contain both x0 and bounds.

True, though it depends whether we want the `IOptimizer` to be exposed as a public interface. If it's completely public, then it shouldn't require anything inside of the `force_bdss` package to run.

PR #352, opened 2020-06-12
scipy algorithms

sparsonslab:
## Description
Make the list of scipy algorithms a global variable so that it can be imported throught the api by other plugins.

## Changes
list of scipy algorithms defined as list object outside of class, rather then as an enum trait inside of the optimizer.


sparsonslab:
I need this into master for my new engine plugin in enthought-example.

Issue #293, opened 2020-02-24
Upper bound attribute for KPIs


Corwinpro:
`KPISpecification` class should implement a `bound` attribute (upper bound for minimization, and lower bound for maximization). This attribute is required by the gradient-free multiobjective optimization algorithms. It also seems quite sensible to implement it in general, as if we have two contradicting objectives, it might be possible that one of those will go to infinity over the course of optimization while the second one is optimized.

flongford:
Closed by #353

PR #353, opened 2020-06-22
ENH: Additional KPI options and caching

flongford:
## Description
Includes additional attributes on `KPISpecification` class and implements caching of raw KPI values in the `BaseOptimizerEngine` class.

### Related Issues
#293 

### Summary
`KPISpecification` class now contains 4 new attributes:
- `target_value`: A value to offset the minimisation routine if `objective == 'TARGET'`
- `use_bounds`: Whether to include upper and lower bounds on the KPI values during the optimisation
- `lower_bound`: A lower bound for the KPI value
- `upper_bound`: An upper bound for the KPI value

`BaseOptimizerEngine` class contains 3 new methods:
- `cache_result`: stores MCO parameter values and corresponding evaluated KPIs
- `retrieve_result`: retrieves KPIs for a given set of MCO parameter values
- `_get_kpi_cache_key`: generates a hashable object from a set of MCO parameter values

Additionally, we enhance the `BaseOptimizerEngine._score` function to both cache the raw KPI values returned from the evaluator, and then return the minimisation score, based on each `KPISpecification.objective` value.

The `IOptimize.optimize_function` and `BaseOptimizerEngine.optimise` method have been slightly amended to expect keyword arguments, which is expected to be useful for passing in any additional values during the MCO.

## Notes
- All spellings of "optimise", "minimise" and "maximise" in the code base have now been converted to American English

## Future Work
- Other FORCE repos (`force_wfmanager`, `force_nevergrad`) should be updated to make use of these new features
- Some testing objects have been changed to accommodate new API and could be tidied up / moved to probe / dummy unit test modules.

## Changes
- New attributes in `KPISpecification` class
- KPI value caching in `BaseOptimizerEngine._score` method
- `convert_to_mask` renamed to `convert_to_score` and can now handle "TARGET" objctives

### DIFF
--- a/force_bdss/mco/optimizer_engines/aposteriori_optimizer_engine.py
+++ b/force_bdss/mco/optimizer_engines/aposteriori_optimizer_engine.py
@@ -37,6 +37,8 @@ class AposterioriOptimizerEngine(BaseOptimizerEngine):
         optimization result: tuple(array-like, array-like)
             MCO parameter and KPI values at point of optimization
         """
+
+        # Clear the KPI cache at the start of the optimization
         self._kpi_cache = {}
 
         #: get pareto set
@@ -45,7 +47,7 @@ class AposterioriOptimizerEngine(BaseOptimizerEngine):
                 self.parameters,
                 **kwargs):
             # Retrieve the cached raw KPI values
-            kpis = self._kpi_cache[tuple(point)]
+            kpis = self.retrieve_result(point)
             yield point, kpis
 
     def unpacked_score(self, *unpacked_input):
--- a/force_bdss/mco/optimizer_engines/aposteriori_optimizer_engine.py
+++ b/force_bdss/mco/optimizer_engines/aposteriori_optimizer_engine.py
@@ -3,7 +3,7 @@
 
 import logging
 
-from traits.api import Str, Instance, Dict
+from traits.api import Str, Instance
 
 from force_bdss.mco.optimizers.i_optimizer import IOptimizer
 
@@ -29,9 +29,6 @@ class AposterioriOptimizerEngine(BaseOptimizerEngine):
     #: callable
     optimizer = Instance(IOptimizer, transient=True)
 
-    #: Caches KPI values between optimization runs
-    _kpi_cache = Dict(transient=True)
-
     def optimize(self, **kwargs):
         """ Generates optimization results.
 
@@ -44,23 +41,12 @@ class AposterioriOptimizerEngine(BaseOptimizerEngine):
 
         #: get pareto set
         for point in self.optimizer.optimize_function(
-                self.objective_function,
+                self._score,
                 self.parameters,
                 **kwargs):
             # Retrieve the cached raw KPI values
             kpis = self._kpi_cache[tuple(point)]
             yield point, kpis
-
-    def objective_function(self, input_values):
-        """From a set of input values, return the objective
-        function for the minimization routine"""
-
-        # Calculate and cache the raw KPI values
-        kpi_values = self._score(input_values)
-        self._kpi_cache[tuple(input_values)] = kpi_values
-
-        # Return the score to be minimized
-        return self._minimization_score(kpi_values)
 
     def unpacked_score(self, *unpacked_input):
         packed_input = list(unpacked_input)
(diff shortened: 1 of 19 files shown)


force_bdss/mco/optimizer_engines/aposteriori_optimizer_engine.py
```diff
@@ -29,19 +29,23 @@ class AposterioriOptimizerEngine(BaseOptimizerEngine):
     #: callable
     optimizer = Instance(IOptimizer, transient=True)
 
-    def optimize(self, *vargs):
+    def optimize(self, **kwargs):
         """ Generates optimization results.
```
sparsonslab, line 34: It might help to have docstrings for the **kwargs.

force_bdss/mco/optimizer_engines/aposteriori_optimizer_engine.py
```diff
@@ -29,19 +29,23 @@ class AposterioriOptimizerEngine(BaseOptimizerEngine):
...
         #: get pareto set
         for point in self.optimizer.optimize_function(
                 self._score,
-                self.parameters):
-            kpis = self._score(point)
+                self.parameters,
+                **kwargs):
+            # Retrieve the cached raw KPI values
+            kpis = self._kpi_cache[tuple(point)]
             yield point, kpis
```
sparsonslab, line 50: So you don't have to worry about self._kpi_cache - its just updates on every _score() which is called by the optimizer?

force_bdss/mco/optimizer_engines/base_optimizer_engine.py
```diff
@@ -76,17 +88,21 @@ def _score(self, input_point):
...
-        score = self.single_point_evaluator.evaluate(input_point)
+
+        # Calculate and cache the raw KPI values
+        kpi_values = self.single_point_evaluator.evaluate(input_point)
+        self._kpi_cache[tuple(input_point)] = kpi_values
+
+        # Return the score to be minimized
+        score = self._minimization_score(kpi_values)
         log.info("Objective score: {}".format(score))
         return score
```
sparsonslab, line 100: This is the heart of it, and as it is never overridden, we don't need to change the code anywhere else (?).

force_bdss/mco/optimizer_engines/aposteriori_optimizer_engine.py
flongford: Yes, you can just refer it during the `optimize` method like here. This way all the BDSS framework objects can be considered in the `BaseOptimizerEngine` subclasses and the `IOptimizer` classes just deal with objects from the wrapper library (i.e. nevergrad)

force_bdss/mco/optimizer_engines/base_optimizer_engine.py
flongford: Yup, anything that sits on top of the `BaseOptimizerEngine` can always access the raw KPI values from the cache. However, the `_score` function returns the actual objective function (with the optimization objectives taken care of). Therefore the external wrappers using the `IOptmizer` interface only need to care about minimizing some function.

flongford:
> I think I get it (see comments below). I guess the nevergrad plugin should still work without any change. Have you tested it?

So running the unit tests only results in some failures because of changing `MINIMISE` to `MINIMIZE` etc. Everything else appears to run ok.

flongford:
> > I think I get it (see comments below). I guess the nevergrad plugin should still work without any change. Have you tested it?
> 
> So running the unit tests only results in some failures because of changing `MINIMISE` to `MINIMIZE` etc. Everything else appears to run ok.

Euurrrggghhh... I just realised *EVERY* example workflow JSON will be broken by this spelling change - I'm going to remove it for now any make a future issue to implement it.

force_bdss/mco/optimizer_engines/aposteriori_optimizer_engine.py
flongford: Has been updated

flongford:
@sparsonslab - merging this should not break anything in the `enthought_example`, `force_nevergrad` and `force_wfmanager` repos.

One unit test needs to be changed in the ITWM repo simply because we reduce the number of times the `Workflow.evaluate` function needs to be called by caching the results: https://github.com/force-h2020/force-bdss-plugin-itwm-example/pull/84

